PR #32, opened 2022-03-14
IRIS-2039-Settings-Header-and-Save-Button

giuliano176:


src/settings/components/general-settings/logout.tsx
```diff
@@ -4,21 +4,27 @@
...
 import { logout } from '../../../network/logout';
+import { settingsSubSections } from '../../../constants';
 
 const Logout: FC = () => {
 	const [t] = useTranslation();
 	const onClick = useCallback(() => {
 		logout();
 	}, []);
+	const sectionTitle = useMemo(
+		() => t(settingsSubSections[6].label, settingsSubSections[6].fallback),
```
geeky-abhishek: I personally will not prefer to read something based on index ,because if in future we add any more subsections then this condition will be disturb ,either used key based approach or use find function

src/settings/components/settings-header.tsx
```diff
@@ -0,0 +1,98 @@
+/*
+ * SPDX-FileCopyrightText: 2021 Zextras <https://www.zextras.com>
```
geeky-abhishek: I guess the license should have year 2022

src/settings/search-settings-view.tsx
```diff
@@ -33,43 +33,62 @@ const SearchSettingsView: FC<{
...
 					label={t(
 						'settings.search_settings.labels.include_search_in_spam_folder',
 						'Include Spam Folder in Searches'
 					)}
-					defaultChecked={searchInSpamFolder}
-					onChange={(e: boolean): void => {
-						setSearchInSpamFolder(e);
-					}}
+					value={searchInSpamFolder}
+					onClick={(): void => onClickSpam()}
```
geeky-abhishek: onClick={onClickSpam} will be better

src/constants/index.ts
```diff
@@ -86,3 +86,34 @@ export const BASENAME = `/${__SHELL_ENV__}/`;
 export const EMAIL_VALIDATION_REGEX =
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars, max-len, no-control-regex
 	/(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/;
+
+export const settingsSubSections = [
```
nubsthead: Shouldn't this be dynamically calculated?

src/settings/components/general-settings/logout.tsx
nubsthead: I also think this is not the proper approach

src/settings/components/general-settings/module-version-settings.tsx
```diff
@@ -37,12 +37,17 @@ const ModuleVersionSettings: FC = () => {
...
+		[t]
+	);
 	return (
 		<>
 			<FormSubSection
-				label={t('module.app.version', 'Application versions')}
+				label={sectionTitle}
 				minWidth="calc(min(100%, 512px))"
 				width="50%"
+				id={sectionTitle.replace(/\s/g, '')}
```
nubsthead: this could be memoized too

src/settings/components/general-settings/out-of-office-view.tsx
```diff
@@ -108,12 +109,16 @@ const OutOfOfficeView: FC<{
...
+		() => t(settingsSubSections[2].label, settingsSubSections[2].fallback),
+		[t]
+	);
 	return (
 		<FormSubSection
-			label={t('settings.out_of_office.headings.settings_label', 'Out of Office Settings')}
+			label={sectionTitle}
 			minWidth="calc(min(100%, 512px))"
 			width="50%"
+			id={sectionTitle.replace(/\s/g, '')}
```
nubsthead: this could be memoized too

src/settings/components/general-settings/user-quota.tsx
```diff
@@ -62,9 +66,10 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 		</Container>
 	) : (
 		<FormSubSection
-			label={t('user_quota.title', "User's quota")}
+			label={sectionTitle}
 			minWidth="calc(min(100%, 512px))"
 			width="50%"
+			id={sectionTitle.replace(/\s/g, '')}
```
nubsthead: this could be memoized too

src/settings/language-and-timezone-settings.tsx
```diff
@@ -55,12 +56,16 @@ const LanguageAndTimeZone: FC<{
...
+		() => t(settingsSubSections[1].label, settingsSubSections[1].fallback),
+		[t]
+	);
 	return (
 		<FormSubSection
-			label={t('settings.timezone_and_language', 'Time Zone and Language')}
+			label={sectionTitle}
 			minWidth="calc(min(100%, 512px))"
 			width="50%"
+			id={sectionTitle.replace(/\s/g, '')}
```
nubsthead: this could be memoized too

src/settings/components/general-settings/module-version-settings.tsx
geeky-abhishek: instead of adding multiple memorised value you can simply add another key with name of id in the settings subsection and use the memo in the following way to find

```
export const settingsSubSections = [
	{
		label: 'settings.general.theme_options',
		fallback: 'Theme Options', 
                 key:'theme_option',
                id:theme_option (or any relevant id)
	}]
const sectionTitle= useMemo(
		() => find(settingsSubSection,{key:"SectionName"}),[dependencies]);
		
```
		
and use it as label: sectionTitle.label ,id:sectionTitle.id something like this

geeky-abhishek (approved):

zovomat (approved):

PR #39, opened 2022-03-18
feat: added tag management

giuliano176:


src/network/tags.ts
```diff
@@ -0,0 +1,40 @@
...
+import { SHELL_APP_ID } from '../constants';
+import { useTagStore } from '../store/tags';
+import { getSoapFetch } from './fetch';
+
+const set = useTagStore.setState;
+export const createTag = (tag: Omit<Tag, 'id'>): Promise<CreateTagResponse> =>
+	getSoapFetch(SHELL_APP_ID)<CreateTagRequest, CreateTagResponse>('CreateTag', {
+		_jsns: 'urn:zimbraMail',
+		tag
+	});
```
geeky-abhishek, line 24: Just a question ,if we need to rename and changeTagColor at the same time ,then do we have to use 2 separate calls renameTag & changeTagColor ? Or can't we create 'updated' action also to modify multiple properties at the same time?

geeky-abhishek (approved):

zovomat (approved):

src/network/tags.ts
zovomat: it makes sense, considering the possibility of an 'edit tag' modal

PR #56, opened 2022-04-01
chore: removed updateTag function

giuliano176:
removed updateTag as the zimbra API is not working


zovomat (approved):

geeky-abhishek:
@zovomat @giuliano176 will it be better to replace it with a batch request ? or should I simply handle it on my side in the code ?

giuliano176:
> 

conceptually I would find it more correct for the apps to call renameTag and changeTagColor in sequence, rather than to implement a "fake" updateTag in the shell

nubsthead (approved):

geeky-abhishek (approved):

PR #92, opened 2022-06-14
feat: palette and icons for shared and linked content

FrancescoGottardi:
Added palette colors and icons for shared & linked items.

### DIFF
--- a/src/boot/theme-provider.tsx
+++ b/src/boot/theme-provider.tsx
@@ -81,6 +81,38 @@ const themeSizes = (
 	}
 };
 
+const paletteExtension =
+	(): ThemeExtension =>
+	(t: any): any => {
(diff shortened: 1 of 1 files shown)


src/boot/theme-provider.tsx
```diff
@@ -81,6 +81,38 @@ const themeSizes = (
 	}
 };
 
+const paletteExtension =
+	(): ThemeExtension =>
+	(t: any): any => {
```
geeky-abhishek: can this 't' cannot be typed ? also can you change the variable name from 't' to something else because we globally use 't' from Translation's T function

dhavaldodiya (approved):

nubsthead (approved):

geeky-abhishek (approved):

PR #94, opened 2022-06-15
feat: code and UI alignment to the design system's new version

gnekoz:


src/reporting/feedback.tsx
```diff
@@ -74,17 +76,21 @@ const TAContainer = styled(Container)`
...
 `;
 
-const LabelContainer = styled(Container)`
-	border-bottom: 1px solid ${(props): string => (props.disabled ? 'red' : '#cfd5dc')};
+interface LabelContainerProps extends ContainerProps {
+	disabled: boolean;
+}
+
+const LabelContainer = styled(Container)<LabelContainerProps>`
+	border-bottom: 1px solid ${({ disabled }): string => (disabled ? 'red' : '#cfd5dc')};
```
geeky-abhishek: You can use theme to get the proper color from design System instead of hardcode values

src/search/search-bar.tsx
```diff
@@ -111,7 +113,11 @@ export const SearchBar: FC<SearchBarProps> = ({
 			if (inputTyped.length > 0) {
 				const newInputState = [
 					...inputState,
-					...map(inputTyped?.split(' '), (item) => ({ label: item, hasAvatar: false }))
+					...map(inputTyped?.split(' '), (item: string, id: number) => ({
```
geeky-abhishek: it can also be define as
`
...map(inputTyped?.split(' '), (label: string, id: number) => ({
						id,
						label,
						hasAvatar: false
					}))
`

src/search/search-bar.tsx
```diff
@@ -347,7 +353,7 @@ export const SearchBar: FC<SearchBarProps> = ({
 				setOptions(suggestions);
 			}
 			return {
-				label: newChip.trim(),
```
geeky-abhishek, line 346: I guess the trimming of chips is already handled in DS ChipInput

gnekoz:
Before merging this PR we need to wait the release of the new version of the DS

src/search/search-bar.tsx
gnekoz: Almost: I used the shorthand for the _label_ property, but the _id_ still need to be converted to string since the new object has to be a _QueryChip_

geeky-abhishek (approved):

dhavaldodiya (approved):

PR #102, opened 2022-07-20
feat: added inline support in composer

geeky-abhishek:
After going through the official Documentation -https://www.tiny.cloud/docs-3x/api/class_tinymce.Editor.html  I didn't find anything regarding the prop **skin_url** but in one blog that only importing the css file is enough and it's already in the import ,also I didn't find any issue in removing it.

### DIFF
--- a/src/store/integrations/composer.tsx
+++ b/src/store/integrations/composer.tsx
@@ -53,15 +53,13 @@ type ComposerProps = {
 	initialValue?: string;
 	/** The content of the editor (controlled mode) */
 	value?: string;
-	/** The base url to append to the resource urls */
-	baseAssetsUrl?: string;
 };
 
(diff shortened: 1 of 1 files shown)


dhavaldodiya (approved):

src/store/integrations/composer.tsx
```diff
@@ -94,20 +94,22 @@ const Composer: FC<ComposerProps> = ({
 			<Editor
 				value={content}
 				init={{
-					skin_url: `${baseAssetsUrl}/tinymce/skins/ui/oxide`,
```
nubsthead, line 97: Can you comment this line for future references?

nubsthead (approved):

gnekoz (approved):

PR #151, opened 2022-11-03
fix: remove font download

giuliano176:


gnekoz (approved):

dhavaldodiya (approved):

beawar (commented):
package-lock.json is not committed

It seems that roboto font is not rendered anymore, in any browser (text are rendered with the fallback font). I think we should do a little more investigation on this

PR #155, opened 2022-11-10
feat: added data-testid for E2E tests

gnekoz:
IRIS-3458


giuliano176 (approved):

src/shell/creation-button.tsx
```diff
@@ -56,6 +56,7 @@ export const CreationButtonComponent: FC<{ activeRoute: AppRoute; location: Loca
 	return primaryAction ? (
 		<Container minWidth="80px">
 			<MultiButton
+				data-testid="NewItemButton"
```
nubsthead, line 59: I think this two should have different ids... how can you test which one of the two component is rendering?

dhavaldodiya (approved):

PR #163, opened 2022-11-18
fix: apply dark mode preview when changed to initial value

CataldoMazzilli:
refs: IRIS-3481

### DIFF
--- a/src/settings/components/general-settings/appearance-settings.tsx
+++ b/src/settings/components/general-settings/appearance-settings.tsx
@@ -8,20 +8,22 @@ import { FormSubSection, Select } from '@zextras/carbonio-design-system';
 import { find } from 'lodash';
 import React, { FC, useCallback, useContext, useMemo } from 'react';
 import { ThemeCallbacksContext } from '../../../boot/theme-provider';
-import { DR_VALUES, SHELL_APP_ID } from '../../../constants';
+import { SHELL_APP_ID } from '../../../constants';
 import { getT } from '../../../store/i18n';
 import { themeSubSection } from '../../general-settings-sub-sections';
 import { useDarkReaderResultValue } from '../../../custom-hooks/useDarkReaderResultValue';
+import { DarkReaderPropValues, isDarkReaderPropValues } from '../../../../types';
 
 const AppearanceSettings: FC<{
 	addMod: (type: 'prefs' | 'props', key: string, value: { value: any; app: string }) => void;
-}> = ({ addMod }) => {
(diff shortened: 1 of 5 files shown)


nubsthead (dismissed):

src/settings/components/general-settings/appearance-settings.tsx
```diff
@@ -43,13 +45,18 @@ const AppearanceSettings: FC<{
 		[darkReaderResultValue, items]
 	);
 	const onSelectionChange = useCallback(
-		(v) => {
-			if (DR_VALUES.includes(v) && v !== darkReaderResultValue) {
-				setDarkReaderState(v);
-				addMod('props', 'zappDarkreaderMode', { app: SHELL_APP_ID, value: v });
+		(value) => {
```
beawar, line 48: value is an implicit any here. We can type the useCallback as `useCallback<NonNullable<SelectProps['onChange']>>`

src/settings/components/general-settings/appearance-settings.tsx
```diff
@@ -43,13 +45,18 @@ const AppearanceSettings: FC<{
...
+			if (isDarkReaderPropValues(value)) {
+				if (value !== darkReaderResultValue) {
+					setDarkReaderState(value);
```
beawar: this set is done twice in both branches of the conditional, just move it out of it :D

src/settings/general-settings.tsx
```diff
@@ -34,6 +34,23 @@ const GeneralSettings: FC = () => {
 			}
 		}));
 	}, []);
+	const removeMod = useCallback((type: 'props' | 'prefs', key) => {
```
beawar: key is an implicit any here. Also, I would type `type` as keyof Mods, but we can discuss about this

src/settings/general-settings.tsx
```diff
@@ -34,6 +34,23 @@ const GeneralSettings: FC = () => {
...
+		setMods((prevState) => {
+			const prevType = prevState[type];
+			if (prevType && prevType[key]) {
+				const nextState = { ...prevState, [type]: { ...prevState[type] } };
+				const nextType = nextState[type];
+				if (nextType && nextType[key]) {
```
beawar: this can create bugs: if `nextType[key]` is indeed present but it is set to 0, false or any other value which is translated to false by js, the next line won't be executed and so the value won't be removed from the object, resulting in weird unwanted behaviours

beawar (approved):

giuliano176 (approved):

PR #188, opened 2023-01-20
Added checked attribute in folders

giuliano176:


src/store/folder/utils.ts
```diff
@@ -15,7 +15,7 @@ const getOriginalId = (f: Folder): string => {
 export const sortFolders = (f: Folder): string => {
 	const id = getOriginalId(f);
 	if (id === FOLDERS.TRASH) {
-		return '~';
+		return '16.1';
```
beawar: What should this value represent? Maybe a comment and/or a constant could be clearer

src/workers/folder.ts
```diff
@@ -187,6 +188,9 @@ export const handleFolderModified = (modified: Array<Partial<UserFolder>>): void
 			const folder = folders[val.id];
 			if (folder) {
 				Object.assign(folder, omit(val));
+				if (typeof val.f !== 'undefined') {
+					folder.checked = /#/.test(val.f ?? '');
```
beawar: this test is used in 2 different ways here and at line 50. Is this wanted? If they do the same (but also if they don't) I would put this test inside a util function because it's not readable as is imo

src/store/folder/utils.ts
giuliano176: this value takes the trash folder in the last position of the system folders. I assigned a constant to it

src/workers/folder.ts
giuliano176: it is wanted, I created a function in utils

gnekoz (approved):

beawar (approved):

gnekoz (approved):

PR #191, opened 2023-01-25
fix: handle management of links when received from notify

giuliano176:


gnekoz (dismissed):

dhavaldodiya (dismissed):

FrancescoGottardi (dismissed):

types/network/soap.d.ts
```diff
@@ -74,6 +74,7 @@ export type SoapNotify = {
 		m?: Array<unknown>;
 		c?: Array<unknown>;
 		folder?: Array<Partial<SoapFolder>>;
+		link?: Array<Partial<SoapFolder>>;
```
beawar: Shouldn't this type be of type SoapLink?
Since the link can be modified, shouldn't it also can be created? Why the link field is not present there?

A more in general consideration: to be honest the SoapNotify type feels like a general type, which could then be extended for the specific worker. I don't know if the current is the right way of setting things up. What if tomorrow we incur in the situation where another module needs to use the workers? Should we add all the specific fields of that module in this type?

beawar (approved):

gnekoz (approved):

PR #193, opened 2023-01-30
fix: avoid loading fonts from ds to prevent slow loading of module

beawar:
Fonts loaded from DS bundle are slowing down a lot the loading of the modules since the request to download the inline font is taking more than 2 seconds to response.

Restore the usage of @fontsource/roboto to load a local copy of the fonts. Update webpack config to place the fonts where fontsource is loading them from.

refs: SHELL-31

### DIFF
--- a/carbonio.webpack.js
+++ b/carbonio.webpack.js
@@ -114,5 +114,15 @@ module.exports = (conf, pkg, options, mode) => {
 		]
 	};
 	conf.externals = {};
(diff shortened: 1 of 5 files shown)


beawar:
- [x] Requires Zextras/carbonio-ui-sdk#9

CataldoMazzilli (approved):

gnekoz (approved):

PR #211, opened 2023-02-24
fix: clear search displayer on clear search

dhavaldodiya:
refs: IRIS-3674

### DIFF
--- a/src/search/search-app-view.tsx
+++ b/src/search/search-app-view.tsx
@@ -19,6 +19,7 @@ import { Redirect, Route, Switch } from 'react-router-dom';
 import { QueryChip, ResultLabelType } from '../../types';
 import AppContextProvider from '../boot/app/app-context-provider';
 import { SEARCH_APP_ID } from '../constants';
+import { pushHistory } from '../history/hooks';
 import { useAppStore } from '../store/app';
 import { getT } from '../store/i18n';
 import { useSearchStore } from './search-store';
@@ -49,11 +50,13 @@ const ResultsHeader: FC<{ label: string; labelType?: ResultLabelType }> = ({
 	const t = getT();
 	const [query, updateQuery] = useQuery();
 	const [, setDisabled] = useDisableSearch();
+	const { module } = useSearchStore();
 
 	const resetQuery = useCallback(() => {
 		updateQuery([]);
 		setDisabled(false);
-	}, [updateQuery, setDisabled]);
+		pushHistory({ route: SEARCH_APP_ID, path: `/${module}` });
+	}, [updateQuery, setDisabled, module]);
 
 	const labelTypeElem = useMemo<ReactElement | undefined>(() => {
 		if (labelType === ResultLabelType.NORMAL) {
--- a/src/search/search-bar.tsx
+++ b/src/search/search-bar.tsx
@@ -25,6 +25,7 @@ import { QueryChip } from '../../types';
 import { getT } from '../store/i18n';
 import { ModuleSelector } from './module-selector';
 import { useSearchStore } from './search-store';
+import { pushHistory } from '../history/hooks';
 
 const OutlinedIconButton = styled(IconButton)`
 	border: 0.0625rem solid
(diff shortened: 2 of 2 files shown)


beawar (changes requested):
I don't think a push history from the shell is the best solution here. With a push history, we are creating a new entry in the browser history, so if you do a lot of searches and clears, and then navigate back with the browser, you'll face a lot of redirect to the search module, with just empty searches and sometimes the displayer populated with something which is not present in the list. This is already happening because of the history.push placed inside the onSearch callback, which is also wrong imo.

Another problem I can see here is the situation where a module need to set some other search param inside the url, and this param hasn't to be cleared with the clear query. With this change, the module cannot accomplish it, since the clear performs a hard reset of the url with a navigation.

I think each module should be in charge of clearing the displayer when needed, and each module should also define what is better between a push and a replace history (a replace could better fit if no additional search param is set in the url, or if the previous search is not saved for a restore).

What the shell can (and should) do, is providing the modules the information of what happens, with all the required additional data updated as well.

dhavaldodiya:
> I don't think a push history from the shell is the best solution here. With a push history, we are creating a new entry in the browser history, so if you do a lot of searches and clears, and then navigate back with the browser, you'll face a lot of redirect to the search module, with just empty searches and sometimes the displayer populated with something which is not present in the list. This is already happening because of the history.push placed inside the onSearch callback, which is also wrong imo.
> 
> Another problem I can see here is the situation where a module need to set some other search param inside the url, and this param hasn't to be cleared with the clear query. With this change, the module cannot accomplish it, since the clear performs a hard reset of the url with a navigation.
> 
> I think each module should be in charge of clearing the displayer when needed, and each module should also define what is better between a push and a replace history (a replace could better fit if no additional search param is set in the url, or if the previous search is not saved for a restore).
> 
> What the shell can (and should) do, is providing the modules the information of what happens, with all the required additional data updated as well.


Hello Beatrice,
Thanks for your feedback! 
We need to manage this bug in our modules but investigating this issue and we found out it's not a bug of only mails module. 
I agree on your proposed solution, How do you suggest to approach the problem? Because I think the shell one should provide the callback to modules. Can we use replaceHistory instead of pushHistory as temporary solution and add a proper solution later?

dhavaldodiya:
> > I don't think a push history from the shell is the best solution here. With a push history, we are creating a new entry in the browser history, so if you do a lot of searches and clears, and then navigate back with the browser, you'll face a lot of redirect to the search module, with just empty searches and sometimes the displayer populated with something which is not present in the list. This is already happening because of the history.push placed inside the onSearch callback, which is also wrong imo.
> > Another problem I can see here is the situation where a module need to set some other search param inside the url, and this param hasn't to be cleared with the clear query. With this change, the module cannot accomplish it, since the clear performs a hard reset of the url with a navigation.
> > I think each module should be in charge of clearing the displayer when needed, and each module should also define what is better between a push and a replace history (a replace could better fit if no additional search param is set in the url, or if the previous search is not saved for a restore).
> > What the shell can (and should) do, is providing the modules the information of what happens, with all the required additional data updated as well.

beawar:
> Hello Beatrice, Thanks for your feedback! We need to manage this bug in our modules but investigating this issue and we found out it's not a bug of only mails module. I agree on your proposed solution, How do you suggest to approach the problem? Because I think the shell one should provide the callback to modules. Can we use replaceHistory instead of pushHistory as temporary solution and add a proper solution later?

@dhavaldodiya Yep I saw the issue is present also in other modules, but I still think is not the best to reset entirely the navigation from here.
And I'm afraid of temporary solutions, because they end up never being replaced.

From the module, you already know if the query has been cleared, because the query array returned by the useQuery prop of the search view will be empty. From there, you can implement your logic to decide what to do.

Another point in favor of letting each module handle this, is that the displayer is not strictly controlled by the url. So a module could have a different way of manage it, and so even the "temporary solution" won't work. Thinking about this, I think this solution is a no-go.

If you find something not working well in the passage of the query, then we can discuss about that and try to improve it or fix it.

beawar:
@dhavaldodiya What do you think of my last considerations? I would like to make this pr go on either by continuing the discussion or by closing it. Thanks

dhavaldodiya:
I am closing it as we will manage it at module side. 
Thank you!

PR #216, opened 2023-03-01
fix: remove tab from suggested separators of search bar

beawar:
refs: SHELL-62

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -169,7 +169,7 @@
         "title": "Primary Account Settings"
     },
     "search": {
-        "active_input_label": "Separate your keywords by a comma or pressing TAB",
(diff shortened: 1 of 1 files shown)


CataldoMazzilli (dismissed):

gnekoz (commented):
Should the instructions also include "space" as separator?

beawar:
> Should the instructions also include "space" as separator?

Good question, I'm checking

beawar:
Checked with the UX team, @gnekoz as you suggested it is better to specify also the space, thank you for pointing that out!
I'm removing the validation label, waiting for an additional validation check.

gnekoz (approved):

CerDennis (approved):

CataldoMazzilli (approved):

PR #217, opened 2023-03-01
refactor: extended the type of the AccountSettings properties

gnekoz:
Extended the type of `prefs` and `attrs` of the `AccountSettings` type to support preferences with array values 

refs: SHELL-66


beawar:
@gnekoz build is failing because of the lint, can you fix it?

beawar (approved):

gnekoz:
> 

@beawar I fixed the issue, thank you for reporting it.
Actually I'm a bit confused because I had that error report during the commit, I fixed it but I forgot to stage the changes and push it. I think it shouldn't have committed the file in the first place. I'll investigate on it

CataldoMazzilli (approved):

PR #232, opened 2023-04-04
Shell 73 fallback tab

CataldoMazzilli:


src/shell/boards/board-tab-list.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+		}));
+	}
+
+	test('If I close the first tab that is open, the tab on its right should be seen correctly', async () => {
+		setupState('board-1');
+		const { user } = setup(<TabsList />);
+		const title1 = screen.getByText('title1');
+		const title2 = screen.getByText('title2');
+		const title3 = screen.getByText('title3');
+		expect(title1).toHaveAttribute('color', 'text');
```
beawar: Might be better to use toHaveStyleRule('color': '#333333') (requires jest-styled-components). I think it's an error that the color attribute is set on the rendered html for the Text component actually, particularly with the "text" value. To avoid having hard coded color, we might create a const available in tests.

src/shell/boards/board-tab-list.tsx
```diff
@@ -39,9 +39,23 @@ export const TabsList: FC = () => {
 				width="calc(100% - 0.5rem)"
 			>
 				{boards &&
-					map(boards, (tab) => (
-						<AppBoardTab key={tab.id} id={tab.id} title={tab.title} icon={tab.icon} />
-					))}
+					map(
+						map(boards, (board) => board),
```
beawar: double map? `map(boards, (board) => board)` can be removed, can't it?

src/shell/boards/board-tab-list.tsx
```diff
@@ -39,9 +39,23 @@ export const TabsList: FC = () => {
...
+							<AppBoardTab
+								key={tab.id}
+								id={tab.id}
+								title={tab.title}
+								icon={tab.icon}
+								fallbackId={
+									current !== tab.id
+										? undefined
+										: (array[index + 1] && array[index + 1].id) ||
+										  (array[index - 1] && array[index - 1].id)
```
beawar: can be shortened in
```ts
current !== tab.id
    ? undefined
    : array[index + 1]?.id || array[index - 1]?.id
```

beawar (approved):

CerDennis (approved):

PR #234, opened 2023-04-07
fix: call noOp request on update view action

beawar:
Revert noop replacement made with a7554b53

refs: SHELL-55


beawar:
Big catch of @CMazzilli while checking code, we were close to introduce a regression here :cold_sweat: .

nubsthead (approved):

CataldoMazzilli (approved):

PR #249, opened 2023-05-19
feat: make board resizable and responsive on default

beawar:
Make board resizable through the drag of the borders.
Save last sizes in local storage and use it when opening new boards.
Update default size to be responsive based on height of the screen and to respect the 4:3 aspect ratio.
Add action to reset board size and position to default one.
Keep board visible for at least one of its borders when resizing the window and a custom size and position was manually set for the board.

refs: SHELL-27 SHELL-97 SHELL-98 SHELL-100 SHELL-102


translations/en.json
```diff
@@ -7,6 +7,10 @@
         "hide": "Hide board",
         "open_app": "Open in app",
         "reduce": "Reduce board",
+        "reset_size": {
+            "disabled": "Board still at the default position",
```
nubsthead: I think the disabled value could improve, the string does not sound right to me

src/shell/boards/resizable-container.tsx
nubsthead, line 1: It's not technically wrong but I think 0.25 rem is too small to be able to click properly on the border to resize the window, I suggest to increase this value to 1rem

nubsthead (approved):

src/shell/boards/resizable-container.tsx
beawar: About this, we first tried with 0.5rem, which was quite usable, giving the user enough "margin of error", but we faced a problem in the right border when a scroll bar is present. The border was placed over the scrollbar, making it difficult to click. So we reduced to 0.25rem. We want to give it a try, see how the users respond, considering it a good compromise.
We surely take this feedback and sum it up with the ones we will receive from the users. Thank you!

src/shell/hooks/useResize.ts
```diff
@@ -0,0 +1,166 @@
...
+	const initialSizeAndPositionRef = useRef<Parameters<typeof calcNewSizeAndPosition>[1]>();
+	const lastSizeAndPositionRef = useRef<Partial<SizeAndPosition>>({});
+	const [lastSavedSizeAndPosition, setLastSavedSizeAndPosition] = useLocalStorage<
+		Partial<SizeAndPosition>
+	>(options?.localStorageKey || 'use-resize-data', {});
+
+	useEffect(() => {
+		if (elementToResizeRef.current) {
+			const elementToResize = elementToResizeRef.current;
+			if (elementToResize) {
```
CataldoMazzilli: Can we remove the following setElementSizeAndPosition calls?

src/shell/boards/board-container.tsx
```diff
@@ -155,62 +201,172 @@ export const BoardContainer: FC = () => {
...
+	const isDefaultSizeAndPos = useMemo(() => size(currentBoardSize) === 0, [currentBoardSize]);
+
+	const resetSizeAndPosition = useCallback(() => {
+		setLastSavedBoardSize({});
+		reduceBoards();
+	}, [setLastSavedBoardSize]);
+
+	useEffect(() => {
+		// reset position when the board is closed
+		if (isEmpty(boards)) {
```
CataldoMazzilli: Maybe we can memoize the isEmpty(boards) info and use it as dependency to improve understanding of effect trigger

CataldoMazzilli (approved):

src/shell/hooks/useResize.ts
beawar: Yeah it is redundant since the style is set from board-container. Thanks, on it

PR #251, opened 2023-05-24
fix: show a modal on failure of getInfo and getComponents

CataldoMazzilli:
refs: SHELL-92


src/boot/bootstrapper-provider.tsx
nubsthead, line 1: I would rename the also file name

src/boot/bootstrapper-router.tsx
```diff
@@ -4,30 +4,24 @@
...
+import { Loader } from './loader';
 
-const ContextBridge: FC = () => {
+const ContextBridge = (): null => {
 	const history = useHistory();
-	// eslint-disable-next-line @typescript-eslint/ban-types
-	const createSnackbar = useContext(SnackbarManagerContext) as Function;
-	// eslint-disable-next-line @typescript-eslint/ban-types
-	const createModal = useContext(ModalManagerContext) as Function;
+	const createSnackbar = useContext(SnackbarManagerContext);
```
beawar: can be replaced with useSnackbar
```suggestion
	const createSnackbar = useSnackbar();
```

src/boot/bootstrapper-router.tsx
```diff
@@ -4,30 +4,24 @@
...
+	const createModal = useContext(ModalManagerContext);
```
beawar: can be replaced with useModal
```suggestion
	const createModal = useModal();
```

src/boot/bootstrapper-router.tsx
```diff
@@ -38,42 +32,42 @@ const ContextBridge: FC = () => {
...
 	useEffect(() => {
 		if (route) useAppStore.setState({ standalone: route });
 	}, [route]);
 	return null;
 };
 
-const DefaultViewsRegister: FC<{ t: TFunction }> = ({ t }) => {
+type DefaultViewsRegisterProps = { t: TFunction };
+
+const DefaultViewsRegister = ({ t }: DefaultViewsRegisterProps): null => {
```
beawar: Probably `t` can be included directly inside this component instead of passing it as prop?

src/boot/bootstrapper-router.tsx
```diff
@@ -38,42 +32,42 @@ const ContextBridge: FC = () => {
...
 	useEffect(() => {
 		registerDefaultViews(t);
 	}, [t]);
 	return null;
 };
 
-const BootstrapperRouter: FC = () => {
+const BootstrapperRouter = (): JSX.Element => {
```
beawar: If we move `t` inside `DefaultViewsRegister`, we could remove this component and move its content inside `Bootstrapper`. It is only used there and there aren't particular dependencies which create rerenders. What do you think?

src/boot/loader.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+jest.mock('../workers');
+jest.mock('../reporting/functions');
+
+describe('Loader', () => {
+	test('If only getComponents request fails, the LoaderFailureModal appears', async () => {
+		// using getInfo and loginConfig default handlers
+		server.use(
+			rest.get<never, never, GetComponentsJsonResponseBody>(
+				'/static/iris/components.json',
+				(req, res, ctx) => res(ctx.status(503))
```
beawar: The test is passing only because there is no valid json in the response. I think we should improve how the fetch is made here
https://github.com/Zextras/carbonio-shell-ui/blob/ef3a88824e75db180b7e26a0e3d872387730078a/src/network/get-components.ts#L12-L14
and check also the status of the response.
See also here https://developer.mozilla.org/en-US/docs/Web/API/fetch, the fetch does not reject if a 404 is returned, but I think we should make the promise throw in this case too.

So, here, the handler could become
```suggestion
			rest.get<never, never, Partial<GetComponentsJsonResponseBody>>(
				'/static/iris/components.json',
				(req, res, ctx) => res(ctx.status(503, 'Controlled error: fail components.json request'), ctx.json({}))
```

while the fetch could be
```ts
fetch('/static/iris/components.json')
    .then((response) => {
        if (response.ok) {
            return response.json();
        }
        throw Error(response.statusText);
    })
```
And inside the jest-env-setup, this line can be removed
https://github.com/Zextras/carbonio-shell-ui/blob/ef3a88824e75db180b7e26a0e3d872387730078a/src/jest-env-setup.ts#L30

src/network/get-info.ts
```diff
@@ -4,44 +4,29 @@
...
-					account,
-					settings,
-					zimbraVersion: version
-				});
-			}
-		})
-		.catch((err: Error) => {
-			console.error(err);
-		});
+	getSoapFetch(SHELL_APP_ID)<{ _jsns: string; rights: string }, GetInfoResponse>('GetInfo', {
```
beawar, line 16: `getSoapFetch` suffers of the same problem of the `getComponents` fetch.
It does not throw if the response has a status different from 200-299 and the body contains a json like 
```json
{
  "Body": {}
}
```

We might consider changing how the getSoapFetch works too

src/boot/loader.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+		setup(<Loader />);
+
+		const title = await screen.findByText('Something went wrong...');
+		act(() => {
+			jest.runOnlyPendingTimers();
+		});
+		expect(title).toBeVisible();
+	});
+
+	test('If only loginConfig request fails, the LoaderFailureModal does not appears', async () => {
```
beawar: typo
```suggestion
	test('If only loginConfig request fails, the LoaderFailureModal does not appear', async () => {
```

src/boot/loader.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+		// using getComponents and getInfo default handlers
+		server.use(rest.post(LOGIN_V3_CONFIG_PATH, (req, res, ctx) => res(ctx.status(503))));
+
+		setup(<Loader />);
+
+		expect(screen.queryByText('Something went wrong...')).not.toBeInTheDocument();
+	});
+
+	test('If Loader requests do not fail, the LoaderFailureModal does not appears', async () => {
```
beawar: typo
```suggestion
	test('If Loader requests do not fail, the LoaderFailureModal does not appear', async () => {
```

src/boot/loader.tsx
```diff
@@ -0,0 +1,93 @@
+/*
+ * SPDX-FileCopyrightText: 2021 Zextras <https://www.zextras.com>
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { Modal, Padding, Text } from '@zextras/carbonio-design-system';
+import React, { FC, useCallback, useEffect, useState } from 'react';
```
beawar: style: unused import
```suggestion
import React, { useCallback, useEffect, useState } from 'react';
```

beawar (dismissed):

beawar (dismissed):

nubsthead (dismissed):

beawar (approved):

CerDennis (approved):

nubsthead (approved):

PR #262, opened 2023-06-06
feat: allow the user to resize and move the board

beawar:
Make board resizable through the drag of the borders.
Save last sizes in local storage and use it when opening new boards.
Update default size to be responsive based on height of the screen and to respect the 4:3 aspect ratio.
Add action to reset board size and position to default one.
Keep board visible for at least one of its borders when resizing the window and a custom size and position was manually set for the board.
Make board movable through the drag of the header
Propagate updates of the local storage to all usages, both in the same and cross window

refs: SHELL-27


CataldoMazzilli (dismissed):

nubsthead (dismissed):

src/shell/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,107 @@
...
+			});
+		}
+	},
+	setValue<T>(key: string, value: React.SetStateAction<T>): void {
+		setState((state) => {
+			const valueToStore = value instanceof Function ? value(state.storage[key]) : value;
+			const newState = { ...state };
+			if (!isSameLocalStorageValue(valueToStore, state.storage[key])) {
+				window.localStorage.setItem(key, JSON.stringify(valueToStore));
+				newState.storage[key] = value;
```
beawar: It is wrongly saving the function instead of the value inside the store if the set is called with the (prevState) => state 
```suggestion
				newState.storage[key] = valueToStore;
```

beawar (commented):
I want to add a couple of tests to the hook to avoid this kind of error in the future

__mocks__/zustand/index.ts
nubsthead, line 1: It is not completely clear to me why you needed to modify this mock

src/shell/hooks/useLocalStorage.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+			</>
+		);
+	};
+
+	test('should show localStorage value and store value', () => {
+		const initial = 'initial';
+		const updated = 'updated';
+		const updatesLocalStorage = [initial];
+		const updatesLocalStorageStore = [undefined, initial];
+		setup(<TestComponent initialValue={initial} updatedValue={updated} />);
```
nubsthead, line 59: Why don't you use renderHook from @testing-library/react-hooks to tests hooks?

beawar: Along with all the articles which uses react-hooks, I've found a couple of articles in which it is suggested to test hooks through simple test components. I found it useful for this case because it allowed me to test also the update of the data across different instances of the hook.

https://github.com/testing-library/react-hooks-testing-library
https://kentcdodds.com/blog/how-to-test-custom-react-hooks

__mocks__/zustand/index.ts
beawar: Long story short: I faced a problem with the initial state having the test-key inside. I thought it was something related with the instances, and I was right.

BUT

You planted a bug in my hear with this ultra appreciated comment, and I checked again the code, debugged a bit, read the documentation again. The problem is not here, it is in the store of the local storage!
In the following 3 blocks I am totally misusing the set state of zustand! :exploding_head: 
I am indeed leaving the same instance of the inner state, I am only creating a new instance of the outer state.

https://github.com/Zextras/carbonio-shell-ui/blob/6c3c7f147b30283a02b9ccfc9ea0239a36c18781/src/shell/hooks/useLocalStorage.ts#L28-L35

https://github.com/Zextras/carbonio-shell-ui/blob/6c3c7f147b30283a02b9ccfc9ea0239a36c18781/src/shell/hooks/useLocalStorage.ts#L38-L42

https://github.com/Zextras/carbonio-shell-ui/blob/6c3c7f147b30283a02b9ccfc9ea0239a36c18781/src/shell/hooks/useLocalStorage.ts#L46-L55


I'm proceeding reverting the changes for this file and fixing this issue. Thank you so much! :fire:

nubsthead: Glad my comment helped. You were also able to anticipate my next questions, well done!

nubsthead (approved):

CerDennis (approved):

CataldoMazzilli (approved):

PR #265, opened 2023-06-19
fix: remove composer default autofocus

FrancescoGottardi:


beawar (commented):
I'm not sure just removing the attribute is the right solution. It is just changing the behaviour for everyone else.
I think it could be better if the editor configs could be instead customized, and the ones included inside the shell would then represent just the defaults.

gnekoz:
> I'm not sure just removing the attribute is the right solution. It is just changing the behaviour for everyone else. I think it could be better if the editor configs could be instead customized, and the ones included inside the shell would then represent just the defaults.

I agree on the config customization but in this case I find weird for a component to take the focus by default: it should be a responsibility of the parent/ancestor component to give focus to one of its inner components.
IMO it could be better to customize the editor configuration in the Mails editor (the only other component that is currently using the Composer, if I'm not wrong) forcing the auto focus, and removing the default value or set it to _false_ in Shell

beawar:
AFAIK the composer is used in calendars and inside mails settings too.
We can of course change the default inside shell to be false, but we should provide a way to set the value from outside, as you said.

About the commit/PR: 
- the suggestion is to use the imperative/present tense for the commit (remove instead of removed)
- the ref is missing, I can see the issue number from the branch name, but it seems unrelated, since it's a issue of contacts, which does not use the composer at all.

PR #268, opened 2023-06-22
fix: fix out of office setting

CataldoMazzilli:
removed all usage of xmlSoapFetch in shell project to avoid encoding-problems with "<>&" characters
deprecated usage of editSetting and removed usage in shell

refs: SHELL-101


src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -104,56 +103,30 @@ const AccountsList = ({
...
+				flgType: 'persona',
+				type: t('label.persona', 'Persona'),
+				identityId: createListrequestId,
+				fromAddress: identities[0]?.fromAddress,
+				identityName: newPersonaName,
+				fromDisplay: identities[0]?.fromDisplay,
+				replyToEnabled: 'FALSE'
+			}
+		]);
 		setCreateListrequestId(createListrequestId + 1);
```
beawar: replace the state createListRequestId with a ref (refactor also the name to be fully camelCase :pray:)

src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -104,56 +103,30 @@ const AccountsList = ({
...
-			saveToSent: '',
-			sentMailFolder: '',
-			whenInFoldersEnabled: '',
-			whenSentToEnabled: ''
-		});
-		setIdentities(identities);
+		setIdentities([
+			...identities,
+			{
+				id: `${identities.length}`,
```
beawar, line 110: I doesn't feel right using the length as id, what if I add one, remove one, and the add one again? The length for the 2 adds will be the same

src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -104,56 +103,30 @@ const AccountsList = ({
 				])
 			) + 1;
```
beawar, line 105: I suggest to open a new issue to improve this code. By seeing it, I'm thinking a reduce could achieve the same result in a clearer and simpler way

src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -104,56 +103,30 @@ const AccountsList = ({
...
 		const newPersonaName = `New Persona ${newPersonaNextNumber || 1}`;
```
beawar, line 106: "New persona" is not translated (could be a different issue)

src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -162,25 +135,11 @@ const AccountsList = ({
 			),
 			(item: IdentityProps, index: number) => ({ ...item, id: index.toString() })
```
beawar, line 137: This piece of code is related to the id being the length of the list. Can be totally avoided by giving a different uniq id to the items (uuid?)

src/settings/components/account-settings/delegates.tsx
```diff
@@ -18,36 +18,60 @@ import {
...
+				const result = reduce(
+					ace,
+					(accumulator: Array<DelegateType>, item, idx) => {
+						const index = findIndex(accumulator, { email: item.d });
+						if (index === -1) {
+							accumulator.push({ email: item.d || '', right: item.right, id: idx.toString() });
+						} else {
+							accumulator.push({
+								email: item.d || '',
+								right: `${item.right} and ${accumulator[index].right}`,
```
beawar: not translated (at least the "and")

src/settings/components/account-settings/delegates.tsx
```diff
@@ -18,36 +18,60 @@ import {
...
+		});
+	}, []);
+
+	const [activeValue, setActiveValue] = useState('1');
 
 	const ListItem = ({ item }: ItemComponentProps<DelegateType>): ReactElement => (
 		<>
 			<Container
-				onClick={(): void => changeView(item.id)}
+				onClick={(): void => setActiveDelegate(item.id)}
```
beawar: move the callback out of the jsx

src/settings/components/account-settings/persona-use-section.tsx
```diff
@@ -54,12 +54,11 @@ const PersonaUseSection = ({
...
-			id: identity.identityId,
-			key: 'zimbraPrefWhenSentToEnabled',
-			value: whenSentToEnabled ? 'FALSE' : 'TRUE'
-		};
-		updateIdentities(modifyProp);
+		updateIdentities(
+			identity.identityId,
+			'zimbraPrefWhenSentToEnabled',
+			whenSentToEnabled ? 'FALSE' : 'TRUE'
+		);
```
beawar: Remove the dependency of the state whenSentToEnabled
```ts
setWhenSentToEnabled((prevState) => {
    const newState = !prevState;
    updateIdentities(
        identity.identityId,
	    'zimbraPrefWhenSentToEnabled',
	    newState ? 'TRUE' : 'FALSE'
    );
    return newState;
});
		
```

src/settings/components/account-settings/settings-sent-messages.tsx
```diff
@@ -83,26 +84,20 @@ const SettingsSentMessages = ({
 
 	const onClickReplyToEnabled = useCallback(() => {
 		setReplyToEnabledValue(!replyToEnabledValue);
```
beawar: remove the dependency from the state replyToEnabledValue by using the prevState of the setState

src/settings/accounts-settings.tsx
```diff
@@ -104,174 +95,126 @@ const getAvailableEmailAddresses = (account: Account, settings: AccountSettings)
...
-					? {
-							...prevState.identity?.createList,
-							[modifyProp.id]: {
-								prefs: { ...prevRecord, [modifyProp.key]: modifyProp.value }
-							}
-					  }
-					: undefined;
-			return {
-				...prevState,
+			const prevRecordPrefs = modifyList[id] || {};
```
beawar: remove the dependency with the state modifyList by using the prevState of the setModifyList

src/settings/accounts-settings.tsx
```diff
@@ -104,174 +95,126 @@ const getAvailableEmailAddresses = (account: Account, settings: AccountSettings)
...
+		(id, key, value) => {
+			if (typeof id === 'string') {
+				updateModifyList(id, key, value);
+			} else {
+				modifyCreateList(id, key, value);
 			}
-			const updatedIdentityKey = lowerFirst(replace(modifyProp.key, 'zimbraPref', ''));
+			const updatedIdentityKey = lowerFirst(replace(key, 'zimbraPref', ''));
 			setIdentities(
 				map(identities, (item) =>
```
beawar: remove the dependency of state identities by using the prevState of setIdentities

src/settings/accounts-settings.tsx
```diff
@@ -104,174 +95,126 @@ const getAvailableEmailAddresses = (account: Account, settings: AccountSettings)
...
-			const arg = {
-				type: 'identity',
-				deleteList
-			};
-			addMod(arg);
+	const removeIdentity = useCallback(
+		(identityId: string | number) => {
+			if (typeof identityId === 'string') {
+				setDeleteList((prevState) => [...prevState, identityId]);
+				if (modifyList[identityId]) {
```
beawar: remove modifyList and createList dependencies by using the prevState of the setState

src/settings/accounts-settings.tsx
```diff
@@ -292,8 +235,136 @@ export const AccountsSettings = ({
...
+				deleteList,
+				(identityId, index): DeleteIdentityRequest => ({
+					_jsns: NameSpace.ZimbraAccount,
+					identity: { id: identityId },
+					requestId: index.toString()
+				})
+			);
+		}
+
+		let modifyIdentityRequests: Array<ModifyIdentityRequest> = [];
```
beawar, line 276: create and modify are almost the same, the code can be unified in a single function which build the common object

src/settings/accounts-settings.tsx
```diff
@@ -292,8 +235,136 @@ export const AccountsSettings = ({
...
+									? reduce(
+											modifyList,
+											(acc, prefs, id) => {
+												const tempResult = [];
+												const propIndex = findIndex(
+													acc,
+													(itemMods, indexAccount) => acc[indexAccount].id === id
+												);
+												if (propIndex > -1) {
+													forEach(Object.keys(prefs), (item, _index) => {
```
beawar: by using Object.entries here instead of cycling over the keys, you could avoid having to call Object.values in the next lines

src/settings/accounts-settings.tsx
```diff
@@ -292,8 +235,136 @@ export const AccountsSettings = ({
...
+						displayName:
+							find(modifyList, (item) => item.zimbraPrefIdentityId === s?.account?.id)
+								?.zimbraPrefIdentityName || s.account?.displayName,
+						identities: {
+							identity:
+								typeof s.account !== 'undefined'
+									? reduce(
+											modifyList,
+											(acc, prefs, id) => {
+												const tempResult = [];
```
beawar: tempResult is not used, can be removed

src/settings/accounts-settings.tsx
```diff
@@ -292,8 +235,136 @@ export const AccountsSettings = ({
...
+											[
+												...filter(
+													s.account.identities.identity,
+													(item) => !deleteList?.includes(item.id)
+												).filter((i) => i.name !== 'DEFAULT'),
+												...map(res?.CreateIdentityResponse, (item) => item.identity[0]),
+												...filter(
+													s.account.identities.identity,
+													(item) => !deleteList?.includes(item.id)
+												).filter((i) => i.name === 'DEFAULT')
```
beawar: double loop on the same s.account.identities.identty on lines 353 and 358, and useless double filter on both lines 355 and 360.
I think we need to create an issue to cleanup this code once for all

src/settings/accounts-settings.tsx
```diff
@@ -318,17 +388,28 @@ export const AccountsSettings = ({
 				}
 				throw new Error(typeof error === 'string' ? error : 'edit setting error');
 			});
+		resetLists();
```
beawar: by resetting lists here, we are not waiting for the success of the operation. I think it would be better if we cleanup the changes only if the operation succeeded, so that (when/if the errors are handled well and the user understand that something went wrong) the user could try again with the same changes, without loosing them

src/settings/general-settings.tsx
```diff
@@ -95,8 +106,81 @@ const GeneralSettings = (): JSX.Element => {
 			return unAppliedPrevState;
 		});
 		if (size(mods) > 0) {
-			const promise = editSettings(mods)
-				.then(() => {
+			let modifyPropertiesRequest: ModifyPropertiesRequest | undefined;
+			if (mods.props) {
```
beawar, line 109: we should avoid using mods state here since we are also setting it

types/account/index.d.ts
```diff
@@ -81,6 +79,13 @@ export interface AccountSettingsPrefs {
 	zimbraPrefIncludeTrashInSearch?: BooleanString;
 	zimbraPrefOutOfOfficeFromDate?: GeneralizedTime;
 	zimbraPrefOutOfOfficeUntilDate?: GeneralizedTime;
+	zimbraPrefHtmlEditorDefaultFontColor?: string;
+	zimbraPrefHtmlEditorDefaultFontFamily?: string;
+	zimbraPrefHtmlEditorDefaultFontSize?: string;
+	zimbraPrefLocale?: string;
+	zimbraPrefMailPollingInterval?: string;
```
beawar: type can be improved. zimbraPrefMailPollingInterval is declared of type duration, which can be translated in
```ts
export type DurationUnit = 'd' | 'h' | 'm' | 's' | 'ms';
export type Duration = `${number}${DurationUnit | ''}`;
```

Then, then cast can be removed from here:
https://github.com/Zextras/carbonio-shell-ui/blob/a9824c0f2441079bf95cebb43a6524437f89b30a/src/store/network/utils.ts#L28

types/account/index.d.ts
```diff
@@ -81,6 +79,13 @@ export interface AccountSettingsPrefs {
...
+	zimbraPrefMailTrustedSenderList?: Array<string> | string;
+	zimbraPrefTimeZoneId?: string;
```
beawar: This one have actually cardinality multi. It could be an array

types/account/index.d.ts
```diff
@@ -95,6 +100,41 @@ export type AccountSettings = {
 	props: Array<ZimletProp>;
 };
 
+export interface IdentityAttrs {
+	/** default mail signature for account/identity/dataSource */
+	zimbraPrefDefaultSignatureId?: string;
+	zimbraPrefForwardReplyFormat?: string;
```
beawar:
```suggestion
	zimbraPrefForwardReplyFormat?: `'text' | 'html' | 'same'`;
```

types/account/index.d.ts
```diff
@@ -95,6 +100,41 @@ export type AccountSettings = {
...
+	/** email address to put in from header.  Deprecated on data source as of bug 67068. */
+	zimbraPrefFromAddress?: string;
+	/** Type of the email address from header. (sendAs or sendOnBehalfOf)  */
+	zimbraPrefFromAddressType: 'sendAs' | 'sendOnBehalfOf';
+	/** personal part of email address put in from header */
+	zimbraPrefFromDisplay?: string;
+	zimbraPrefIdentityId?: string;
+	/** name of the identity */
+	zimbraPrefIdentityName?: string;
+	zimbraPrefMailSignatureStyle?: string;
```
beawar:
```suggestion
	zimbraPrefMailSignatureStyle?: 'outlook' | 'internet';
```

types/network/index.d.ts
```diff
@@ -176,3 +185,135 @@ export type TagActionResponse = {
 	action: { op: string; id: string };
 	_jsns: string;
 };
+
+type NameOrIdRequired =
```
beawar: The type defined as it is is not actually useful. The never rule is overwritten by the last definition, with both the fields required.

Also, it is convenient to create a type util to achieve this behaviour instead of replicating the conditions
```ts
type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyof T, Keys>> &
	{
		[K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>;
	}[Keys];
```
to be used as follow
```ts
type ModifyIdentityRequest = {
	_jsns: NameSpace.ZimbraAccount;
	identity: {
		_attrs?: Partial<IdentityAttrs>;
	} & RequireAtLeastOne<Pick<Identity, 'id' | 'name'>>;
};
```

types/network/index.d.ts
```diff
@@ -176,3 +185,135 @@ export type TagActionResponse = {
...
+		right: Right;
+		// Type:String
+		// Zimbra ID of the grantee
+		zid?: string;
+	}>;
+};
+
+export type Right =
+	// Following rights are partial, they are the result of
+	// zmprov gar -v  -t account -c USER
```
beawar: I would remove the command from here, or use the carbonio one.

types/network/index.d.ts
```diff
@@ -176,3 +185,135 @@ export type TagActionResponse = {
...
+	// right class: USER
+	| 'sendOnBehalfOf'
+	// description: view free/busy
+	// right type: preset
+	// target type(s): account
+	// grant target type: (default)
+	// right class: USER
+	| 'viewFreeBusy';
+
+export enum NameSpace {
```
beawar: enums should not be in the declaration file

https://lukasbehal.com/2017-05-22-enums-in-declaration-files/

types/network/index.d.ts
```diff
@@ -176,3 +185,135 @@ export type TagActionResponse = {
...
+}
+
+// The type of grantee:
+// usr - Zimbra user
+// grp - Zimbra group(distribution list)
+// all - all authenticated users
+// gst - non-Zimbra email address and password (not yet supported)
+// key - external user with an accesskey
+// pub - public authenticated and unauthenticated access
+export type GranteeType =
```
beawar, line 284: the comment for each entry could be placed near the value. Also, some items have not a description

src/settings/accounts-settings.tsx
```diff
@@ -104,174 +95,126 @@ const getAvailableEmailAddresses = (account: Account, settings: AccountSettings)
...
-						modifyList
-					}
-				};
-			});
-		},
-		[setMods]
-	);
+	const onCancel = useCallback(() => {
+		resetLists();
+		setIdentities(identitiesDefault);
```
beawar, line 122: on cancel, if a new identity was added but not save, this identity is removed from the list, but the selected identity is not updated, and all sections except the list of identities disappear.

It could be wanted, but I think we should align the behaviour with the one of the delete, where instead the previous identity is selected as fallback.

src/settings/accounts-settings.tsx
beawar: This is also causing the save/discard button to become disabled even if the operation fails

src/settings/general-settings.tsx
beawar: Created SHELL-124 to cleanup all unnecessary states

src/settings/components/account-settings/delegates.tsx
CataldoMazzilli: Created SHELL-125

src/settings/components/account-settings/accounts-list.tsx
CataldoMazzilli: Created SHELL-125

src/settings/components/account-settings/accounts-list.tsx
CataldoMazzilli: Created SHELL-126

src/settings/components/account-settings/accounts-list.tsx
CataldoMazzilli: Created SHELL-127

src/settings/components/account-settings/accounts-list.tsx
CataldoMazzilli: Created SHELL-127

src/settings/accounts-settings.test.tsx
```diff
@@ -0,0 +1,756 @@
...
+import { identityToIdentityProps } from './account-wrapper';
+import server, { waitForRequest } from '../mocks/server';
+
+jest.mock('../workers');
+
+describe('Account setting', () => {
+	test('Show primary identity inside the list', async () => {
+		const firstName = faker.person.firstName();
+		const lastName = faker.person.lastName();
+		const fullName = faker.person.fullName({ firstName, lastName });
```
beawar: firstName and lastName are not required, you can use fullName and email directly with no args. If you want to have all this fields valued with the same name, create a util to populate an identity, so that it can be reused across all these tests and more.

src/settings/accounts-settings.test.tsx
```diff
@@ -0,0 +1,756 @@
...
+			rights: { targets: [] },
+			signatures: { signature: [] },
+			id,
+			displayName: '',
+			identities: {
+				identity: identitiesArray
+			}
+		};
+
+		const { getAllByRoleWithIcon, user } = setup(
```
beawar: cleanup unused things

src/settings/accounts-settings.test.tsx
```diff
@@ -0,0 +1,756 @@
...
+			<AccountsSettings account={account} identitiesDefault={identitiesDefault} />
+		);
+		await screen.findByText('sendAs');
+		const persona1Row = screen.getByText(persona1FullName);
+		expect(persona1Row).toBeVisible();
+		await user.click(persona1Row);
+		await user.click(screen.getByRole('button', { name: /delete/i }));
+		const confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
+		await user.click(confirmButton);
+		expect(persona1Row).not.toBeInTheDocument();
```
beawar, line 312: There is a missing await here which is causing the test suites to fail randomly based on timers.
When you delete the persona1, the default identity will be then selected, causing a second render of the delegates section, and so the effect run, the async network request is run etc..
```suggestion
        await screen.findBytext('sendAs');
		expect(persona1Row).not.toBeInTheDocument();
```

src/settings/accounts-settings.test.tsx
```diff
@@ -0,0 +1,756 @@
...
+			<AccountsSettings account={account} identitiesDefault={identitiesDefault} />
+		);
+		await screen.findByText('sendAs');
+
+		await user.click(screen.getByRole('button', { name: /add persona/i }));
+		const persona1Row = screen.getByText('New Persona 1');
+		expect(persona1Row).toBeVisible();
+
+		await user.click(screen.getByRole('button', { name: /discard changes/i }));
+		expect(persona1Row).not.toBeInTheDocument();
```
beawar, line 493: Also here, persona1 is added and automatically selected, so the delegates section is unmounted. Then on discard, persona1 is removed from the list and default identity is selected again, an the delegates section is rendered again, network request, etc.
```suggestion
        await screen.findByText('sendAs');
		expect(persona1Row).not.toBeInTheDocument();
```

src/settings/accounts-settings.test.tsx
```diff
@@ -0,0 +1,756 @@
...
+		await user.click(screen.getByRole('button', { name: /delete/i }));
+		let confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
+		await user.click(confirmButton);
+		expect(persona1Row).not.toBeInTheDocument();
+
+		await user.click(persona2Row);
+		await user.click(screen.getByRole('button', { name: /delete/i }));
+		confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
+		await user.click(confirmButton);
+		expect(persona2Row).not.toBeInTheDocument();
```
beawar: :warning: This test is not behaving as you might think. Try replacing line 389 to run the expect with the string value instead of the element (screen.getByText('New Persona 1')) and you wiil see the test fails. This is because persona1Row is no more in the document after the creation of the second identity! Try by putting an expect(persona1Row).toBeInTheDocument() right after line 388.

So basically you are not clicking on anything on lines 394 and 400, and you are deleting persona 3 first and persona 2 after that, because of the automatic selection of the previous item. And the expect of line 398 and 404 are successful because the elements are not in the document since before.

Finally, you need to be sure the entire ui is completely updated before doing other stuff, so you need to wait for the other sections to update too when clicking on and deleting items. And when you delete persona1, the default is selected automatically, delegates appears and getRightsRequest is fired. And then also for persona2, click on it will unmount again the delegates section, and the delete will then select automatically the default again, delegates section is mounted again and a new getRightsRequest is fired. We need to wait for all of this inside this test. 

```suggestion
		await user.click(screen.getByRole('button', { name: /add persona/i }));
		await waitFor(() =>
			expect(screen.getByRole('textbox', { name: /persona name/i })).toHaveDisplayValue(
				/new persona 1/i
			)
		);

		const persona1 = 'New Persona 1';
		const persona2 = 'New Persona 2';
		const persona3 = 'New Persona 3';
		expect(screen.getByText(persona1)).toBeVisible();

		await user.click(screen.getByRole('button', { name: /add persona/i }));
		await waitFor(() =>
			expect(screen.getByRole('textbox', { name: /persona name/i })).toHaveDisplayValue(persona2)
		);
		expect(screen.getByText(persona2)).toBeVisible();

		await user.click(screen.getByRole('button', { name: /add persona/i }));
		await waitFor(() =>
			expect(screen.getByRole('textbox', { name: /persona name/i })).toHaveDisplayValue(persona3)
		);
		expect(screen.getByText(persona3)).toBeVisible();

		await user.click(screen.getByText(persona1));
		await waitFor(() =>
			expect(screen.getByRole('textbox', { name: /persona name/i })).toHaveDisplayValue(persona1)
		);
		await user.click(screen.getByRole('button', { name: /delete/i }));
		let confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
		await user.click(confirmButton);
		await screen.findByText(/primary account settings/i);
		await screen.findByText('sendAs');
		expect(screen.queryByText(persona1)).not.toBeInTheDocument();

		await user.click(screen.getByText(persona2));
		await waitFor(() =>
			expect(screen.getByRole('textbox', { name: /persona name/i })).toHaveDisplayValue(persona2)
		);
		await user.click(screen.getByRole('button', { name: /delete/i }));
		confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
		await user.click(confirmButton);
		await screen.findByText(/primary account settings/i);
		await screen.findByText('sendAs');
		expect(screen.queryByText(persona2)).not.toBeInTheDocument();

		await user.click(screen.getByRole('button', { name: /save/i }));
```

src/settings/accounts-settings.tsx
```diff
@@ -9,55 +9,44 @@ import { Container, useSnackbar } from '@zextras/carbonio-design-system';
 import type { TFunction } from 'i18next';
```
beawar: Unused import
```suggestion
```

src/settings/general-settings.tsx
```diff
@@ -5,11 +5,21 @@
...
+import {
+	AccountState,
+	AddMod,
+	BatchRequest,
+	ModifyPrefsRequest,
+	ModifyPrefsResponse,
+	ModifyPropertiesRequest,
+	ModifyPropertiesResponse,
+	Mods,
+	NameSpace,
```
beawar: unused import
```suggestion
```

beawar (approved):
I think we can consider this finished (it is already enormously out of scope from the initial issue).
All the not-fixed problems not related to this PR have now a specific issue or an open point of discussion inside SHELL-116.

nubsthead (approved):

PR #271, opened 2023-06-29
Shell 109 delegates setting not working

CataldoMazzilli:


src/settings/accounts-settings.test.tsx
beawar, line 1: discard test can be added as a test.todo, waiting for the refactor of the radio, on order to check that the discard reset the value to the initial one

src/settings/components/account-settings/delegates.tsx
```diff
@@ -20,23 +20,59 @@ import {
...
-	const [activeValue, setActiveValue] = useState('1');
+	const [activeValue, setActiveValue] =
+		useState<AccountSettingsPrefs['zimbraPrefDelegatedSendSaveTarget']>(delegatedSendSaveTarget);
+
+	const radioGroupOnChange = useCallback<(newValue: string) => void>(
+		(newValue) => {
+			function isZimbraPrefDelegatedSendSaveTarget(
+				arg: string
+			): arg is NonNullable<AccountSettingsPrefs['zimbraPrefDelegatedSendSaveTarget']> {
+				return ['owner', 'sender', 'both', 'none'].includes(arg);
```
beawar: we could extract this array of all possible values in a global const (like DELEGATED_SEND_SAVE_TARGET = ['owner', 'sender', 'both', 'none']).
and then we could use the array to type the pref inside the types, as typeof DELEGATED_SEND_SAVE_TARGET[number].
An finally we could use the same const to check the type assertion here. Doing so we are sure that we are checking for all and only the possible values of the pref, reducing the error of checking inside mistyped strings.

I'm thinking about this because if I add 'something' as value of this array of line 58, nothing tells me it's not a valid value. By introducing a const, we will have the type build from the same array we use across the codebase, and so there is a single point of failure (the constant array)

src/settings/accounts-settings.tsx
```diff
@@ -338,6 +361,15 @@ export const AccountsSettings = ({
 					}
 					return {
 						...s,
+						settings: {
+							...s.settings,
+							prefs: {
+								...s.settings.prefs,
+								zimbraPrefDelegatedSendSaveTarget: delegatedSendSaveTarget as NonNullable<
```
beawar: instead of casting to a nonnullable, we could set empty string as fallback, which is translated to a "reset to default". From the current state of the UI, it should not be possible set this value, but I think that by doing so we can consider it a "covered" case.

src/settings/components/account-settings/delegates.tsx
nubsthead: I agree wuth this comment solution

src/settings/accounts-settings.test.tsx
```diff
@@ -432,10 +423,6 @@ describe('Account setting', () => {
 		);
 		await user.click(screen.getByRole('button', { name: /delete/i }));
 		confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
```
CataldoMazzilli:
```suggestion
		confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
		act(() => {
			// run modal timers
			jest.runOnlyPendingTimers();
		});
```

src/settings/accounts-settings.test.tsx
```diff
@@ -418,10 +413,6 @@ describe('Account setting', () => {
 		);
 		await user.click(screen.getByRole('button', { name: /delete/i }));
 		let confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
```
CataldoMazzilli:
```suggestion
		let confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
		act(() => {
			// run modal timers
			jest.runOnlyPendingTimers();
		});
```

src/settings/accounts-settings.test.tsx
```diff
@@ -313,10 +312,6 @@ describe('Account setting', () => {
 		await user.click(persona1Row);
 		await user.click(screen.getByRole('button', { name: /delete/i }));
 		const confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
```
CataldoMazzilli:
```suggestion
		const confirmButton = await screen.findByRole('button', { name: /delete permanently/i });
		act(() => {
			// run modal timers
			jest.runOnlyPendingTimers();
		});
```

nubsthead (dismissed):
I am still not a big fan of the values with hardcoded strings inside the delegates Radios. I would have used a different approach on declaring the constant to use it both as values and as type.

What I would have done is:

`export const DELEGATED_SEND_SAVE_TARGET = {
	OWNER: 'owner',
	SENDER: 'sender',
	BOTH: 'both',
	NONE: 'none'
} as const;

// utils to create a type from object values
type ObjectValues<T> = T[keyof T];

// actuale type exported around
export type DelegatedSendSaveTarget = ObjectValues<typeof DELEGATED_SEND_SAVE_TARGET>;

In this way you have the constant which can be used as single value inside the radios( so you can do DELEGATED_SEND_SAVE_TARGET.OWNER instead of hardcoding 'owner') , all the possible values (values(DELEGATED_SEND_SAVE_TARGET)) and the type which is automatically updated everytime anyone add a new key to the object
`

I still think your implementation works perfectly fine so i'll leave it as potential suggestion for another/future use case

gnekoz (dismissed):

giuliano176 (dismissed):

lisaparma (dismissed):

beawar (approved):

lisaparma (approved):

gnekoz (approved):

PR #282, opened 2023-07-07
fix: set max 100 as value in the quota component in general settings

rodleyorosa:
refs: SHELL-123

### DIFF
--- a/src/settings/components/general-settings/user-quota.tsx
+++ b/src/settings/components/general-settings/user-quota.tsx
@@ -58,7 +58,7 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 	return mobileView ? (
 		<Container width="fit" padding={{ right: 'medium' }}>
 			<Tooltip label={description} placement="bottom">
-				<Quota fill={quota > 0 ? quota : 0} fillBackground={fillBackground} />
+				<Quota fill={quota >= 100 ? 100 : quota} fillBackground={fillBackground} />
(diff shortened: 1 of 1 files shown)


src/settings/components/general-settings/user-quota.tsx
```diff
@@ -58,7 +58,7 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 	return mobileView ? (
 		<Container width="fit" padding={{ right: 'medium' }}>
 			<Tooltip label={description} placement="bottom">
-				<Quota fill={quota > 0 ? quota : 0} fillBackground={fillBackground} />
+				<Quota fill={quota >= 100 ? 100 : quota} fillBackground={fillBackground} />
```
beawar: This logic does not correspond to the one below. Here you are not considering the -1 value.
I think we should merge the two conditionals (this one and the one of line 79) in a single one, and use the resulting const in the props.
Check if the quota can have value -1 or not before deciding which of the two logic we need to keep

rodleyorosa: I've merged the two conditionals with a useMemo() called filledQuota ad called it in the props. Is it right?

beawar: yep, seems good now

nubsthead (approved):

beawar (approved):

CataldoMazzilli (approved):

PR #284, opened 2023-07-11
feat(composer): add a new prop to set initialization custom options

gnekoz:
- Add a new prop to set initialization custom options

refs: SHELL-122

### DIFF
--- a/src/store/integrations/composer.tsx
+++ b/src/store/integrations/composer.tsx
@@ -57,6 +57,7 @@ type ComposerProps = EditorProps & {
 	 * If defined, a menu item to add inline images is added to the composer.
 	 */
 	onFileSelect?: (arg: { editor: TinyMCE; files: HTMLInputElement['files'] | undefined }) => void;
+	customInitOptions?: EditorProps['init']
 };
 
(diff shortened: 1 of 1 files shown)


src/store/integrations/composer.tsx
```diff
@@ -57,6 +57,7 @@ type ComposerProps = EditorProps & {
 	 * If defined, a menu item to add inline images is added to the composer.
 	 */
 	onFileSelect?: (arg: { editor: TinyMCE; files: HTMLInputElement['files'] | undefined }) => void;
+	customInitOptions?: EditorProps['init'];
```
beawar: I think it could be safer if the type was marked as Partial, even if at the moment all fields are optional inside that init field of the EditorProps
```suggestion
	customInitOptions?: Partial<EditorProps['init']>;
```

beawar (dismissed):

giuliano176 (dismissed):

giuliano176 (approved):

beawar (approved):

lisaparma (approved):

CataldoMazzilli (approved):

PR #293, opened 2023-07-24
refactor: remove IdentityProps layer and account-wrapper

CataldoMazzilli:
refs: SHELL-127


src/settings/components/account-settings/accounts-list.tsx
```diff
@@ -169,6 +105,57 @@ const AccountsList = ({
...
+									<Container width="fit" mainAlignment="flex-end">
+										<Text weight="regular" size="small">
+											{isPrimary(item)
+												? t('label.primary', 'Primary')
+												: t('label.persona', 'Persona')}
+										</Text>
+									</Container>
+								</Row>
+
+								<Row width="fit"></Row>
```
beawar: can be removed?

src/settings/components/account-settings/persona-settings.tsx
```diff
@@ -7,33 +7,42 @@
...
 
 	const onChange = useCallback(
 		(ev) => {
 			setPersonaValue(ev.target.value);
-			updateIdentities(items.identityId, 'zimbraPrefIdentityName', ev.target.value);
+			if (identityAttrs.zimbraPrefIdentityId) {
+				updateIdentities(
+					identityAttrs.zimbraPrefIdentityId,
+					'zimbraPrefIdentityName',
+					ev.target.value
```
beawar, line 41: we probably should type the callback, ev is an implicit any here

src/settings/components/account-settings/settings-sent-messages.tsx
```diff
@@ -64,42 +67,54 @@ const SettingsSentMessages = ({
...
+		);
+	}, [fromAddressArray, identityAttrs.zimbraPrefFromAddress]);
 	useEffect(() => {
-		setReplyToDisplay(identity?.replyToDisplay === undefined ? '' : identity?.replyToDisplay);
-	}, [identity?.replyToDisplay]);
+		setReplyToDisplay(
+			identityAttrs.zimbraPrefReplyToDisplay === undefined
+				? ''
+				: identityAttrs.zimbraPrefReplyToDisplay
+		);
```
beawar: This one can be shortened to
```suggestion
		setReplyToDisplay(identityAttrs.zimbraPrefReplyToDisplay ?? '');
```

src/settings/components/utils.ts
```diff
@@ -1244,3 +1253,97 @@ export function upsertPrefOnUnsavedChanges(
...
+	const filteredAndModified = reduce(
+		modifiedIdentitiesAttrs,
+		(accumulator, attrs, id) => {
+			const propIndex = findIndex(accumulator, (identity) => identity.id === id);
+			if (propIndex > -1) {
+				accumulator[propIndex]._attrs = {
+					...accumulator[propIndex]._attrs,
+					...attrs
+				};
+				if (attrs.zimbraPrefIdentityName && accumulator[propIndex].name !== 'DEFAULT') {
```
beawar: I really like the isPrimary util
```suggestion
				if (attrs.zimbraPrefIdentityName && !isPrimary(accumulator[propIndex])) {
```

src/settings/accounts-settings.test.tsx
```diff
@@ -24,17 +23,391 @@ describe('Account setting', () => {
...
+			...previousState,
+			account
+		}));
+
+		setup(<AccountsSettings />);
+
+		await waitForGetRightsRequest();
+		const renderedItems = screen.getAllByRole('listitem');
+		expect(renderedItems.length).toEqual(4);
+		expect(head(renderedItems.map((item) => item.textContent))).toMatchInlineSnapshot(
```
beawar, line 344: Why not using the toHaveTextContent? I feel it as easier to read.
Even if you need to check multiple entries, with a `getAllBy().every((item, index) => expect(item).toHaveTextContent(expected[index]))`

src/settings/accounts-settings.tsx
```diff
@@ -37,194 +32,181 @@ import {
...
 	const settings = useUserSettings();
 
-	const [delegatedSendSaveTarget, setDelegatedSendSaveTarget] = useState<
-		AccountSettingsPrefs['zimbraPrefDelegatedSendSaveTarget']
-	>(settings.prefs.zimbraPrefDelegatedSendSaveTarget);
+	const createRecord = useRef<Record<string, IdentityAttrs>>({});
+	const deleteArray = useRef<Array<string>>([]);
+	const modifyRecord = useRef<Record<string, Partial<IdentityAttrs>>>({});
+
+	const delegatedSendSaveTarget = useRef<AccountSettingsPrefs['zimbraPrefDelegatedSendSaveTarget']>(
```
beawar: Should the ref variables have the Ref suffix in their name? Btw nice job the removal of the useless states :heart_eyes:

beawar (commented):
On the UI there is a weird scroll inside the ListV2 of the accounts. It is because the List is shrinked a bit vertically. Unfortunately setting the prop flexShrink={0} does not work because of a bug of the DS (just opened CDS-176). Probably waiting for the fix, we could create a temporary styled component (or set an inline style, even if I don't like this option so much)

src/settings/accounts-settings.test.tsx
CataldoMazzilli: https://youtu.be/M9S54fcBF3g Here he suggest to use toMatchInlineSnapshots

beawar (approved):

nubsthead (approved):

lisaparma (approved):

PR #296, opened 2023-07-25
feat: avoid to call getRightsRequest on every mount

rodleyorosa:
refs: SHELL-128


src/settings/accounts-settings.test.tsx
```diff
@@ -1290,5 +1295,88 @@ describe('Account setting', () => {
...
+					(req, res, context) => {
+						requestFn();
+						return getRightsRequest(req, res, context);
+					}
+				)
+			);
+
+			const TestComponent = (): JSX.Element => (
+				<>
+					<AccountsSettings account={account} identitiesDefault={identitiesDefault} />
```
beawar: To really be sure the request is only called at first render, you should place the account settings inside a <Route>, inside the Switch. By doing so, when navigating inside the the other page, the accounts settings  will unmount, to be then mounted again on the second navigation

types/account/index.d.ts
```diff
@@ -15,6 +15,20 @@ export interface ZimletProp {
 	_content: string;
 }
 
+export type GetRightsResponse = {
```
beawar: Not sure about the place of this type. @CMazzilli since you also managed the types lately, where do you think we should place this type to avoid the import cycle?

beawar (dismissed):

CataldoMazzilli (dismissed):

CataldoMazzilli (approved):

beawar (approved):

PR #312, opened 2023-08-17
fix(search): translate module inside search bar placeholder

beawar:
refs: SHELL-87


CataldoMazzilli (dismissed):

src/search/search-bar.tsx
```diff
@@ -63,8 +64,19 @@ export const SearchBar = (): JSX.Element => {
...
+	const {
+		updateQuery,
+		module: currentSearchModuleRoute,
+		query,
+		searchDisabled,
+		setSearchDisabled,
+		tooltip
+	} = useSearchStore();
+	const modules = useAppStore((s) => s.views.search);
+	const module = useMemo(
```
nubsthead: It seems like this module is calculated just for the label at line 298. Am I missing something? does it make sense to do it just for that?

giuliano176 (dismissed):

src/search/search-bar.tsx
beawar: We could extract and memoize the label directly. I did a first wrong implementation were I was using module also for the other checks (replacing the currentSearchModuleRoute), but I reverted this change since it was breaking things.

We need the translated label of the module, and for now the only place where we can find it without a more complex refactor, is inside the search views of the appStore

nubsthead (approved):

CataldoMazzilli (approved):

PR #314, opened 2023-08-18
fix: avoid routes to be overwritten when set in the store

beawar:
Sometimes the module are not shown in the primary bar even if they are loaded right. By logging the draft of the addRoute producer, it seems that the order of execution of the addRoute, plus the direct set made by the default views, make the primary views array length change in an unpredictable way. Sometimes happens that the length in the execution n is greater than the length in the execution n+1.

This fix aims to avoid this kind of race condition, by registering all the routes (so even the default ones) through the setters provided by the store. In the setters, the curried produce of immer is used to update the store through a direct modification of the draftState. By doing so, there should not be present anymore the situation where the state is entirely replaced by an older version of the draft, since the draft is always incrementally updated with the push (to add) and the splice (to remove), and never replaced with a direct assignation.

refs: SHELL-152


beawar:
I've created the PR towards release, but it is better to merge it in devel first to follow the usual flow

src/store/app/store.ts
```diff
@@ -168,147 +171,138 @@ export const useAppStore = create<AppState>()((set, get) => ({
...
 			const { standalone } = get();
 			if (standalone && data.route !== standalone) {
 				return STANDALONE_RESPONSE;
 			}
 			set(
-				produce((state: AppState) => {
-					state.views.settings = sortBy(unionBy([data], state.views.settings, 'id'), 'position');
+				produce<AppState>((state) => {
+					if (!some(state.views.settings, (view) => view.id === data.id)) {
+						addAndSort(state.views.settings, data);
```
CataldoMazzilli: is not already checked inside addAndSort?

beawar: yep I forgot to remove it

src/store/app/store.ts
```diff
@@ -62,103 +71,97 @@ export const useAppStore = create<AppState>()((set, get) => ({
...
+							}
+							accumulator.appContexts[app.name] = {};
+						}
+						return accumulator;
+					},
+					{ moduleApps: {}, shellApp: {}, appContexts: {} }
+				);
+				return {
+					apps: moduleApps,
+					shell: {
```
CataldoMazzilli: shellApp contains all default values so we can simplify with shell: shellApp ?

src/store/app/store.ts
```diff
@@ -62,103 +71,97 @@ export const useAppStore = create<AppState>()((set, get) => ({
 		secondaryBarAccessories: []
 	},
 	setters: {
-		addApps: (apps: Array<Partial<CarbonioModule>>): void => {
-			set((state) => ({
-				apps: reduce(
+		addApps: (apps): void => {
```
CataldoMazzilli: I suggest to rename 'addApps' in 'setApps | setupApps | initApps'

CataldoMazzilli (dismissed):

CataldoMazzilli (approved):

nubsthead (approved):

PR #315, opened 2023-08-22
fix: fix search bar bug that grows in width

rodleyorosa:
refs: SHELL-49


rodleyorosa:
I simply used the vw (viewport width) to manage the sizing of this element in relation to the width of the browser window.

beawar (approved):
By removing the maxWidth in the shell-header, we are removing the reserved space for the utility actions. I'm not crazy about it. But I think we need to review the spaces and widths of the entire header, and so I'm going to accept this for now

CataldoMazzilli (approved):

PR #328, opened 2023-09-18
refactor: replace custom badge with ds badge

rodleyorosa:
refs: SHELL-103


gnekoz (dismissed):

types/apps/index.d.ts
```diff
@@ -48,7 +48,7 @@ export type AppRouteData = AppRoute & {
 
 export type BadgeInfo = {
 	show: boolean;
-	count?: number;
+	count: number | string;
```
beawar: I don't feel good about making this field required (which implies a breaking change btw).

Also, if we allow the string type, the commit should be marked as feat, since we are adding a new feature

src/shell/badge-wrap.tsx
```diff
@@ -5,47 +5,33 @@
...
-					) : null}
-				</MiniBadge>
-			)}
+			{badge.show && <MiniBadge value={badge.count} type={'unread'} />}
 			{children}
 		</Container>
 	)
 );
 
+BadgeWrap.displayName = 'BadgeWrap';
```
beawar: Conventionally with the other components/project, we prefer using a named function inside the forwardRef instead of declaring the displayName directly.
Something like
```ts
const BadgeWrap = React.forwardRef<HTMLDivElement, { badge: BadgeInfo }>(
	function({ badge, children }, ref) {
		return (
           ...
```

src/shell/badge-wrap.tsx
```diff
@@ -5,47 +5,33 @@
...
+	padding: 0.125rem;
 `;
 
-// eslint-disable-next-line react/display-name
 const BadgeWrap: FC<{ badge: BadgeInfo }> = forwardRef<HTMLDivElement, { badge: BadgeInfo }>(
 	({ badge, children }, ref) => (
 		<Container width={'3rem'} height={'3rem'} style={{ position: 'relative' }} ref={ref}>
-			{badge.show && (
-				<MiniBadge badge={badge} height="fit" width="fit">
-					{badge.showCount ? (
```
CataldoMazzilli, line 35: be aware that here you are doing a BC removing the use of badge.showCount;
I think we should avoid a BC

src/shell/badge-wrap.tsx
```diff
@@ -5,47 +5,33 @@
...
-							size="extrasmall"
-							style={{ padding: '0.125rem 0.25rem', fontSize: '0.625rem' }}
-							color="gray6"
-						>
-							{badge.count ?? 0}
-						</Text>
-					) : null}
-				</MiniBadge>
-			)}
+			{badge.show && <MiniBadge value={badge.count} type={'unread'} />}
```
beawar: Also badge.color is not used anymore, by setting "unread" as fixed value the badge will always be colored with primary color. I think we need to keep the possibility to set the custom color, by passing a $color prop to the styled component (instead of the type prop)

src/shell/badge-wrap.tsx
```diff
@@ -5,47 +5,34 @@
...
-					) : null}
-				</MiniBadge>
-			)}
-			{children}
-		</Container>
-	)
+	function BadgeWrapFn({ badge, children }, ref): JSX.Element {
+		return (
+			<Container width={'3rem'} height={'3rem'} style={{ position: 'relative' }} ref={ref}>
+				{badge.show && <MiniBadge value={badge.showCount && badge.count ? badge.count : 0} />}
```
beawar: There is still a problem here: in the previous implementation, if showCount is false, no counter is shown, just an empty badge. While in the current if showCOunt is fals,e, 0 is shown

rodleyorosa: @beawar actually value prop of MiniBadge requires number | string type, is it ok to set it to empty string instead of 0?

rodleyorosa: something like `<MiniBadge value={badge.showCount ? badge.count ?? 0 : ''} />`

beawar: Yes, exactly, there are 3 possible outcomes here, not just 2

beawar (commented):
The implementation seems ok to me now. I think we should add a minimum set of tests to this component to avoid future regressions

src/shell/shell-primary-bar.test.tsx
```diff
@@ -446,4 +446,90 @@ describe('Shell primary bar', () => {
 
 		expect(queryByRoleWithIcon('button', { icon: ICONS.settings })).not.toBeInTheDocument();
 	});
+
+	describe('Primary Badge', () => {
+		test('should render the counter badge if show and showCount are true and count is greater than or equal to 0', () => {
```
beawar: Checking the code again, I think we should just say that the counter badge is shown if count is valued. Negative numbers are accepted too.

beawar (dismissed):

beawar (approved):

nubsthead (approved):

CataldoMazzilli (approved):

PR #332, opened 2023-09-28
refactor: use logout instead of loginOp

CataldoMazzilli:
refs: SHELL-149


src/mocks/handlers.ts
```diff
@@ -21,7 +22,8 @@ const handlers: RequestHandler[] = [
 	rest.post('/service/soap/EndSessionRequest', endSessionRequest),
 	rest.get(LOGIN_V3_CONFIG_PATH, getLoginConfig),
 	rest.get('/', rootHandler),
```
beawar: this one can be cleanup since it was created just for the /?loginOp

beawar (approved):

rodleyorosa (approved):

PR #346, opened 2023-10-17
fix: discard changes of timezone, locale and dark mode reset initial …

CataldoMazzilli:
…values

refs: SHELL-161

### DIFF
--- a/src/settings/language-and-timezone-settings.tsx
+++ b/src/settings/language-and-timezone-settings.tsx
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-import React, { FC, useCallback, useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 
 import {
 	Container,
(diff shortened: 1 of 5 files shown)


nubsthead (dismissed):

src/settings/language-and-timezone-settings.tsx
```diff
@@ -4,74 +4,95 @@
...
-	const defaultLocale = useMemo(
-		() =>
-			(settings.prefs.zimbraPrefLocale &&
-				find(locales, { id: `${settings.prefs.zimbraPrefLocale}` })) ||
-			find(locales, { id: 'en' }),
-		[locales, settings.prefs.zimbraPrefLocale]
+	const [prefLocale, setPrefLocale] = useState<string>(settings.prefs.zimbraPrefLocale ?? '');
+
+	const prefLocaleSelectedValue = useMemo<SelectItem>(
+		() => find(locales, (item) => item.value === prefLocale) as SelectItem,
```
beawar, line 56: Can we add a comment here with the ref of the issue with which we will remove the cast?

src/settings/language-and-timezone-settings.tsx
```diff
@@ -4,74 +4,95 @@
...
-			(settings.prefs.zimbraPrefTimeZoneId &&
-				find(timezones, { value: `${settings.prefs.zimbraPrefTimeZoneId}` })) ||
-			find(timezones, { value: 'UTC' }),
-		[timezones, settings.prefs.zimbraPrefTimeZoneId]
+	const [prefTimeZoneId, setPrefTimeZoneId] = useState<string>(
+		settings.prefs.zimbraPrefTimeZoneId ?? ''
+	);
+
+	const prefTimeZoneIdSelectedValue = useMemo<SelectItem>(
+		() => find(timezones, (item) => item.value === prefTimeZoneId) as SelectItem,
```
beawar, line 76: Can we add a comment with the issue ref?

src/settings/language-and-timezone-settings.tsx
```diff
@@ -80,31 +101,28 @@ const LanguageAndTimeZone: FC<{
...
+					background={'gray5'}
+					label={t('label.language', 'Language')}
+					onChange={onLocaleChange}
+					selection={prefLocaleSelectedValue}
+					showCheckbox={false}
+					dropdownMaxHeight="12.5rem"
+					selectedBackgroundColor="highlight"
+				/>
+
+				<Padding top="small" width="100%">
```
beawar: we could remove this padding and use the container gap instead

beawar (commented):
Discard and save buttons remain enabled also when, after changing some selection, I reset everything to the previous situation.
I've checked on account settings and I think we never defined how the UI should behave in this situation, because some settings are disabling the buttons once the user set everything as it was, some others are leaving the buttons enabled. I feel we need to establish a single common behavior

CataldoMazzilli:
> Discard and save buttons remain enabled also when, after changing some selection, I reset everything to the previous situation. I've checked on account settings and I think we never defined how the UI should behave in this situation, because some settings are disabling the buttons once the user set everything as it was, some others are leaving the buttons enabled. I feel we need to establish a single common behavior

ok created specific issue

beawar (approved):

nubsthead (approved):

nubsthead (approved):

PR #367, opened 2023-12-04
fix: tag worker is invoked only when notify relevant to tags

giuliano176:


src/workers/tag.ts
```diff
@@ -17,36 +17,60 @@ export const handleTagRefresh = (tags: Array<Tag>): Tags => {
...
 	);
 
+function deletedIdsContainTags({
+	deletedIds,
+	tagIds
+}: {
+	deletedIds: Array<string>;
+	tagIds?: Array<string>;
+}): boolean {
+	return deletedIds.some((deletedId) => tagIds?.includes(deletedId));
```
gnekoz, line 62: I think this optimization check is correct but should be placed inside the `handleTagDeleted` after line 41, since the function already receives all the data it needs to do the check.
In this way all the check regarding the deletion will be incapsulated in just one place and the `onmessage` function, which role is just to dispatch the changes, will be simpler

giuliano176: Unfortunately that will not work. Once the postmessage is invoked a value will be returned, the store will be updated and the component will rerender. This is the current situation already.
This fix prevents the update of the store by calling the postmessage only when changes are related to the tags.

dhavaldodiya (approved):

FrancescoGottardi (approved):

nubsthead (approved):

PR #381, opened 2024-02-19
fix: show quota when used quota is 0

rodleyorosa:
Refs: SHELL-184

### DIFF
--- a/src/settings/components/general-settings/user-quota.tsx
+++ b/src/settings/components/general-settings/user-quota.tsx
@@ -24,7 +24,7 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 	const used = useAccountStore((s) => s.usedQuota);
 	const quota = useMemo(() => {
 		const userQuota = Number(settings?.attrs?.zimbraMailQuota);
-		if (used && userQuota && userQuota > 0) {
+		if ((used || used === 0) && userQuota && userQuota > 0) {
 			return Math.round((used / userQuota) * 100);
 		}
 		return -1;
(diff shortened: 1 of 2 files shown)


src/settings/components/general-settings/user-quota.tsx
```diff
@@ -39,7 +39,7 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 
 	const description = useMemo(() => {
 		switch (true) {
-			case !quota || quota < 0:
+			case quota < 0:
 				return t('user_quota.unlimited', 'You have unlimited space available');
 			case quota === 100:
```
nubsthead, line 44: I believe this should be >= 100

rodleyorosa: I'll wait other reviews first

src/settings/components/general-settings/user-quota.tsx
```diff
@@ -24,7 +24,7 @@ const UserQuota: FC<UserQuotaProps> = ({ mobileView }) => {
 	const used = useAccountStore((s) => s.usedQuota);
 	const quota = useMemo(() => {
 		const userQuota = Number(settings?.attrs?.zimbraMailQuota);
-		if (used && userQuota && userQuota > 0) {
+		if ((used || used === 0) && userQuota && userQuota > 0) {
```
beawar: `used` is always a number, so this check is redundant. Also, the userQuota check is doubled imo

src/settings/components/general-settings/user-quota.tsx
beawar: A >= check is for sure an improvement, since right now the message is shown only if the used quota is exactly 100%. But I think it is a change of behavior, even if it is so small. I would prefer having a separate issue and PR. Just my opinion, of course.

beawar (approved):

CataldoMazzilli (approved):

src/settings/components/general-settings/user-quota.tsx
rodleyorosa: I created an issue separated: https://zextras.atlassian.net/browse/SHELL-200

PR #382, opened 2024-02-26
fix: prevent overwriting display of main route if focus mode is true

rodleyorosa:
Refs: SHELL-188

### DIFF
--- a/src/settings/components/general-settings/module-version-settings.test.tsx
+++ b/src/settings/components/general-settings/module-version-settings.test.tsx
@@ -1,0 +1,146 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Zextras <https://www.zextras.com>
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import React from 'react';
+
+import { act } from '@testing-library/react';
+import { produce } from 'immer';
+
+import ModuleVersionSettings from './module-version-settings';
+import { useAppStore } from '../../../store/app';
+import { screen, setup } from '../../../test/utils';
+
+describe('Module Version Settings', () => {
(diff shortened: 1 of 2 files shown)


src/settings/components/general-settings/module-version-settings.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+		expect(screen.getByText(/active/i)).toBeVisible();
+	});
+
+	it('should not render the carbonio shell application version', () => {
+		const description = 'The Zextras Carbonio web client';
+		const display = 'Shell';
+		const version = '5.1.0';
+		setup(<ModuleVersionSettings />);
+		act(() => {
+			useAppStore.getState().setters.setApps([
```
beawar: Is there a reason why the setter of the state is made after the setup? If not, I think we should set the data before rendering the component

lisaparma (dismissed):

gnekoz (dismissed):

src/settings/components/general-settings/module-version-settings.test.tsx
```diff
@@ -44,7 +44,6 @@ describe('Module Version Settings', () => {
 		const description = 'The Zextras Carbonio web client';
 		const display = 'Shell';
 		const version = '5.1.0';
-		setup(<ModuleVersionSettings />);
 		act(() => {
```
beawar: act is not needed now

beawar (approved):
Opened SHELL-199 for a related problem when registering multiple routes for the same module (not in focus mode). We prefer treating this one as a separate issue since it requires some reasoning on it

lisaparma (approved):

CataldoMazzilli (approved):

PR #383, opened 2024-02-29
build!: use node 18

beawar:
fix: update exported types to remove anys
refactor: remove casts and anys and improve internal types
build: update all deps
test: update tests
chore: improve pre-commit script
chore: add missing licenses and update thirdparties

BREAKING CHANGE: use node 18
Refs: SHELL-191


src/store/boards/hooks.tsx
```diff
@@ -43,5 +41,5 @@ export const BoardProvider: FC<{ id: string }> = ({ children, id }) => {
 	);
 };
 
-export const useBoardHooks = (): BoardHooksContext => useContext(boardHooksContext);
-export const useBoard = (): Board => useContext(boardContext);
+export const useBoardHooks = (): BoardHooksContext | undefined => useContext(boardHooksContext);
```
CataldoMazzilli, line 44: should we update also the exposed type and mark it as BC?

beawar: I am not sure if it is better to expose the possibility to have undefined, or have an assertion inside the hook.
The assertion creates a "logical" requirement (which is already present): the board hooks should be called only inside boards.
Exposing the undefined on the other hand would allow the invocation of the hook from everywhere, even if it is useless.

nubsthead (dismissed):

nubsthead (approved):

CerDennis (approved):

beawar:
- [x] Create the last release with node 16 before merging this PR

CataldoMazzilli (approved):

src/store/boards/hooks.tsx
beawar: With @CMazzilli we decided that the fix on the type exported will be managed with SHELL-157, which is specific for the cleanup and re-organization of the types

PR #389, opened 2024-03-15
fix!: cleanup and re-organize npm library

beawar:
The goal of this PR is to align the types of the npm package with the objects injected at runtime.

To do so, all the runtime objects are exported in a specific entry point for the library (`src/lib.ts`) and the declarations of the types are created by typescript itself, removing the need to keeping them updated manually.
The match between the library objects and the runtime ones is double checked by making the runtime injection satisfies the type of the library itself. In this way, a compilation error will occur if there is a difference between the two.

This is a first step to remove the need of mocking the shell entirely when testing the modules. There are still some open points which are preventing this:
- workers are created using import.meta syntax, which is not compatible with current ts configs
- notification.mp3 asset is missing inside the library: a quick fix could be including all the assets or only it, but I need more time to investigate the best approach
- all functions that are app-dependant are at the moment declarations only, and their source code is not packed in the library. We need to understand how to handle them.

### Breaking changes

Even if most of the changes represents a refactor, and should be compatible with the previous version, there are some points that will break projects.

#### AppSetters type is no longer exported

Replace it with the construct `typeof <function>` 

#### _types_ folder is no longer in the library

If you have some import from subpath of the shell module, remove it and use the root import only. E.g.
```ts
// this kind of import is no longer supported
import { something } from '@zextras/carbonio-shell-ui/types/and/other';

// replace it with
import { something } from '@zextras/carbonio-shell-ui';
```

#### Declarations of inexisting code are removed

Inside manually created types, there were some declarations of some objects, fields or functions which did not exist in the source code. These declarations are now no longer there. Their usage will now cause a compilation error, preventing unwanted bugs on runtime.

#### Refactor JSNS enum to an object and improve network types

Soap types have been a little improved, and the type of the __jsns_ field has been further restricted to always match the Namespace type. Since generic strings can no longer be assigned to this field, the object JSNS, which was previously wrongly exported as an enum inside declarations, is now available as an object and can be used to value the field.

### Other changes

- Update _main_ and _types_ inside the package.json to reflect the structure of the library, and the scripts to build both the production package and the library.
- Introduce api-extractor to keep track of the api model. It should reduce the risk of releasing breaking change, since any change to it will probably represent a breaking change for the library.
- To organize better imports and avoid dependency cycles, introduce a couple of rules in eslint, which will split and distinguish all import of the types.
- Deprecate all constants that are not part of shell, and add a note to export to module only those constants which are useful, not the internal ones.
- Improve some internal types, and refactor some constant to use lowercase fields inside objects.
- Export and use JSNS for request namespaces, and improve Soap types.
- Remove some dead code.
- Remove some ts-ignore and fix the affected code.

BREAKING-CHANGE: Replace entirely the declarations of the library. See #389 for the list of changes.
Refs: SHELL-157


beawar:
- [x] Requires: zextras/jenkins-zapp-lib#13

src/settings/accounts-settings.tsx
```diff
@@ -7,52 +7,61 @@
...
-import { useAccountStore, useUserAccount, useUserSettings } from '../store/account';
+	ModifyIdentityRequest,
+	ModifyIdentityResponse,
+	ModifyPrefsRequest,
+	ModifyPrefsResponse,
+	ModifyPropertiesRequest
+} from '../types/network';
+import type { AccountACEInfo } from '../types/network/entities';
+
+export type AccountsSettingsBatchRequest = BatchRequest<{
```
CataldoMazzilli, line 50: It is used only in 1 test file...do we need to export this?

src/store/i18n/store.ts
```diff
@@ -4,17 +4,30 @@
...
 import { dropRight, forEach, reduce } from 'lodash';
 import { initReactI18next } from 'react-i18next';
 import { create } from 'zustand';
 
-import type { CarbonioModule, I18nState } from '../../../types';
 import { SHELL_APP_ID } from '../../constants';
+import type { CarbonioModule } from '../../types/apps';
 import { useAccountStore } from '../account';
 
+export type I18nState = {
```
CataldoMazzilli, line 19: Can we flat and split also this store type?

src/store/integrations/getters.tsx
```diff
@@ -4,32 +4,28 @@
...
 };
 
-export const getIntegratedComponent = (id: string): [FunctionComponent<unknown>, boolean] => {
+export const getIntegratedComponent = (
+	id: string
+): [React.ComponentType<Record<string, unknown>>, boolean] => {
 	const Integration = useIntegrationsStore.getState().components?.[id];
 	if (Integration) {
-		const C: FC = (props: unknown) => (
+		const C = (props: Record<string, unknown>): React.JSX.Element => (
```
CataldoMazzilli: Can we rename C?

src/settings/accounts-settings.tsx
beawar: What's your idea?
I noticed that I should have used this type also in the fetch made at line 303, as a replacement for the current BatchRequest type
https://github.com/zextras/carbonio-shell-ui/blob/db1ba8e23d6229630cb17445c62c064b159b73a3/src/settings/accounts-settings.tsx#L302-L309

package.json
```diff
@@ -100,12 +103,12 @@
 	},
 	"dependencies": {
 		"@fontsource/roboto": "^5.0.8",
-		"@reduxjs/toolkit": "^1.9.7",
+		"@reduxjs/toolkit": "^1 >=1.9",
```
CataldoMazzilli: restore previous version

src/boot/app/load-app.ts
```diff
@@ -4,41 +4,45 @@
...
-					...getAppSetters(appPkg),
-					...getAppFunctions(appPkg),
+					...getAppDependantSetters(appPkg),
+					...otherSetters,
+					...getAppDependantFunctions(appPkg),
+					...otherFunctions,
 					...CONSTANTS,
 					Tracker
-				};
+				} satisfies typeof ExportsForApp;
```
CataldoMazzilli, line 45: I like it

src/settings/accounts-settings.tsx
beawar: @CMazzilli check the refactored code for these types. I've improved it a bit in order to have the Batch type only accept keys suffixed with `Request` or `Response`.
By doing so, an error emerged in the way the type was used in the test

CataldoMazzilli (dismissed):

nubsthead (dismissed):

CataldoMazzilli (approved):

nubsthead (approved):

PR #402, opened 2024-04-10
fix: module boards can trigger mousedown events

nubsthead:
Refs: SHELL-175

### DIFF
--- a/src/shell/boards/board-container.test.tsx
+++ b/src/shell/boards/board-container.test.tsx
@@ -692,6 +692,7 @@ describe('Board container', () => {
 			jest.advanceTimersToNextTimer();
 		});
 		const board = screen.getByTestId(TESTID_SELECTORS.board);
+		const elementForMove = screen.getByTestId(TESTID_SELECTORS.boardHeader);
 		const boardInitialSizeAndPos = buildBoardSizeAndPosition();
 		const boardNewPosition = {
 			top: 0,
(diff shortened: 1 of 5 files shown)


beawar (approved):
I went through the history of the move and even if in the original requirement the move was to make it trigger from the entire board, in the same first implementation it emerged that only the header bar was actually triggering the move, since it was the only component outside the app-board content.

We should ask for a validation round, specific for the move, but for the rest the changes look good to me :rocket:

nubsthead:
> I went through the history of the move and even if in the original requirement the move was to make it trigger from the entire board, in the same first implementation it emerged that only the header bar was actually triggering the move, since it was the only component outside the app-board content.
> 
> We should ask for a validation round, specific for the move, but for the rest the changes look good to me 🚀

Do you want me to ask for a validation for this specific PR?

beawar:
> Do you want me to ask for a validation for this specific PR?
 
@nubsthead yeah I would feel safer to specify to the validator that the move action is now fully restricted to the header component, even if this should not change anything on the user side.

nubsthead:
> > Do you want me to ask for a validation for this specific PR?
> 
> @nubsthead yeah I would feel safer to specify to the validator that the move action is now fully restricted to the header component, even if this should not change anything on the user side.

@beawar The fact is I restricted it to the header just to keep the same behaviour it already has. It can be enabled to the entire board moving back the moveHandler function on its previous component. I moved it to avoid changing its current behaviour.

However I will ask a validation to make sure it is what they expected

CataldoMazzilli (approved):

PR #403, opened 2024-04-12
feat: use json format v4 for translations

beawar:
Update json format to v4.
Remove translations subtree and scripts.
Remove babel-plugin-i18next-extract.
From now on, new labels should be added directly in weblate,
so there is no need to extract them from the code.

Refs: SHEL-208

---

Requirements:
- [x] Weblate project is configured to use json format v4
- [x] Translations for 24.5 are ended and pulled
- [x] Update sdk to use zextras/carbonio-ui-sdk#37


beawar:
- [ ] Weblate project is configured to use json format v4

CataldoMazzilli (dismissed):

CataldoMazzilli (dismissed):

src/i18next.d.ts
nubsthead, line 1: should the import in line 10 be removed?

beawar: Yes, good catch

nubsthead (approved):

CataldoMazzilli (approved):

PR #410, opened 2024-04-16
fix: update module when the module in store is undefined

rodleyorosa:
Added some missing tests inside `search-bar.tsx` component

Refs: SHELL-209


rodleyorosa:
- [x] Add missing tests in the search bar component

beawar (commented):
I'm not entirely convinced about the behavior of the search. There is a redirect that is made only one time when modules are loading, which makes the search always use only the first module that load (and not the first module overall). To give an example, I'm in /search/files; I refresh the page; calendars is the first module loaded; the search redirects itself to calendar search view.

I tried to play a bit with the code, and I saw that it might be not possible to avoid the redirect right now, because we don't know if all modules are loaded yet (and so we don't know if the path is a real module or not, we might need to rethink the search flow entirely), but I think we should at least redirect to the first module overall. In other words, I think we should have a consistent behavior: the search always redirects to the first module (based on priority order), and not the randomly first module to be loaded.

Furthermore, because the redirect is made while modules are still loading, the path logic is useless, because the initial path probably won't match the first module, and so there will be the redirect which changes the path. It could be simplified a lot more by just doing a redirect to the first module, without checking the path at all.

src/search/module-selector.tsx
```diff
@@ -68,10 +60,10 @@ const ModuleSelectorComponent = ({ app }: ModuleSelectorProps): React.JSX.Elemen
 		) {
 			updateModule((modules.find((m) => m.app === app) ?? modules[0])?.route);
 		} else if (module === undefined && app === SEARCH_APP_ID) {
-			updateModule(pathModuleWithSearch);
-			pushHistory({ route: SEARCH_APP_ID, path: `/${pathModuleWithSearch}` });
+			updateModule(modules[0]?.route);
+			pushHistory({ route: SEARCH_APP_ID, path: `/${modules[0]?.route}` });
```
beawar: module[0] here is still the first module loaded, not the first module based on the order/priority

src/search/search-bar.test.tsx
```diff
@@ -351,17 +352,155 @@ describe('Search bar', () => {
...
+			route: app1.name,
+			component: (): React.JSX.Element => <div>{app1.name}</div>
+		} satisfies SearchView;
+		useAppStore.getState().addSearchView(app1SearchView);
+		sessionStorage.setItem(SEARCH_MODULE_KEY, app1.name);
+		setup(<SearchBar />);
+		const selector = screen.getByTestId(TESTID_SELECTORS.headerModuleSelector);
+		expect(within(selector).getByText(app1.display)).toBeVisible();
+		expect(screen.getByRole('textbox', { name: `Search in ${app1.display}` })).toBeVisible();
+	});
```
beawar: The test seems too simple, it will succeed even if you remove the set of the session storage. Additionaly, there is only 1 searchView registered, so in any case it will be the one shown. Add at least another searchView and set inside the storage the app with the greatest priority.

src/search/search-bar.test.tsx
```diff
@@ -351,17 +352,155 @@ describe('Search bar', () => {
...
+			id: app2.name,
+			route: app2.name
+		});
+		setupAppStore([app1, app2], [app1Route, app2Route]);
+		const app1SearchView = {
+			app: app1.name,
+			icon: app1.icon,
+			id: app1.name,
+			label: app1.display,
+			position: 6,
```
beawar: Give to the searchView the same position of the app to keep the coherence with the runtime data

src/search/search-bar.test.tsx
```diff
@@ -458,16 +458,27 @@ describe('Search bar', () => {
...
+			icon: app2.icon,
+			id: app2.name,
+			label: app2.display,
+			position: app2.priority,
+			route: app2.name,
+			component: (): React.JSX.Element => <div>{app2.name}</div>
+		} satisfies SearchView;
+		useAppStore.setState((state) => ({
+			views: { ...state.views, search: [app1SearchView, app2SearchView] }
+		}));
```
beawar: Instead of doing this, set the storage before registering the searchViews (as it should be in the scenario we are testing, the session storage value is already set when the modules register their search views). Also in the other tests the order of the lines should be updated.

beawar (approved):

CataldoMazzilli (approved):

PR #419, opened 2024-05-03
fix: remove incomplete delegates management

gnekoz:
- Removed the Delegates component, and all the related code, from the account settings page
- Removed the Delegates component file
- Updated the AccountsSettings component tests

Refs: SHELL-214


src/settings/accounts-settings.tsx
beawar, line 1: I think also the part of the delegatedSendSaveTargetRef and the zimbraPrefDelegatedSendSaveTarget can be removed. The value of delegatedSendSaveTargetRef is never updated and zimbraPrefDelegatedSendSaveTarget were previously modified through the deleted section.

src/settings/accounts-settings.tsx
```diff
@@ -284,16 +235,6 @@ export const AccountsSettings = (): React.JSX.Element => {
 
 		let modifyPrefsRequest: ModifyPrefsRequest | undefined;
```
beawar: Now this is not used anymore. Consider also to clean the type AccountSettingsBatchRequest/Response, which is specific for this setting and can be cleaned.

beawar (dismissed):

nubsthead (dismissed):

src/settings/accounts-settings.tsx
```diff
@@ -360,53 +292,19 @@ export const AccountsSettings = (): React.JSX.Element => {
...
-				_jsns: JSNS.account,
-				ace: [{ right: 'sendAs' }, { right: 'sendOnBehalfOf' }]
-			}).then((value) => {
-				if (value.ace) {
-					const { ace } = value;
-					setDelegates(generateDelegateList(ace, t));
-				}
-				useAccountStore.setState((state) => ({
-					...state,
-					rights: value.ace ?? []
```
CataldoMazzilli, line 402: this 'rights' field can be removed from the store and relative type

CataldoMazzilli:
Also src/mocks/handlers/getRightsRequest.ts   can be removed now (also the usage in src/mocks/handlers.ts )

beawar:
@gnekoz do you need some help with this?

src/settings/accounts-settings.tsx
beawar: @gnekoz 
https://github.com/zextras/carbonio-shell-ui/blob/425338dc0598747d2d33fa865e374cbb600540b2/src/store/account/store.ts#L15
this is the field @CMazzilli was referring to. I don't see it removed from the new changes.
It was added to prevent requesting the delegates on each rerender, so it is not used anymore.

beawar:
ping @gnekoz

CataldoMazzilli (approved):

rodleyorosa (approved):

beawar (approved):

PR #420, opened 2024-05-06
Update translations 20240506

beawar:


beawar:
There aren't new translations to pull

PR #423, opened 2024-05-20
fix: call /zx/auth/logout on Carbonio

CataldoMazzilli:
Refs: SHELL-206


src/utility-bar/bar.test.tsx
```diff
@@ -84,6 +91,46 @@ describe('Shell utility bar', () => {
...
+
+		const zxAuthLogoutHandlerMock = jest.fn(zxAuthLogoutHandler);
+		server.use(http.get('/zx/auth/logout', zxAuthLogoutHandlerMock));
+
+		useLoginConfigStore.setState((s) => ({ ...s, isCarbonioCE: true }));
+		const { user } = setup(<ShellUtilityBar />);
+
+		await user.click(screen.getByRoleWithIcon('button', { icon: ICONS.accountUtilityMenu }));
+		await user.click(screen.getByText(/logout/i));
+		await waitForRequest('get', '/logout');
```
beawar: I think it is safer to invoke the waitForRequest before the setup, and then wait for the returned promise when needed. This because the waitForRequest search for the requests that comes after it has been invoked. Placing the invocation here might be a risk, considering that timings are not always the same across different executions.

beawar (dismissed):

src/store/login/hooks.test.tsx
nubsthead, line 1: Just a curiosity. Why did you prefer testing a component instead of testing directly the hook itself?

nubsthead (dismissed):

src/store/login/hooks.test.tsx
CerDennis: +1

CerDennis (dismissed):

src/store/login/hooks.test.tsx
CataldoMazzilli: It is the same.
Simply because in this project is used this way until now

nubsthead (approved):

CerDennis (approved):

PR #432, opened 2024-06-06
fix: add link in toolbar and use link instead of quicklink

frisonisland:
Found display: flex breaks quicklink menu visualization.
Added "link" item in toolbar to allow inserting a link

Refs: CO-1146


src/store/integrations/composer.tsx
```diff
@@ -256,8 +257,7 @@ const Composer: FC<ComposerProps> = ({
 		<Container
 			height="100%"
 			crossAlignment="baseline"
-			mainAlignment="flex-start"
-			style={{ overflowY: 'hidden' }}
+			style={{ overflowY: 'hidden', display: 'initial' }}
```
beawar: Instead of using a display: initial in a DS Container, which is meant to be a flex, and exposes all the flex properties, why not using a div directly?
The additional style could be applied with a StyledComponent, if needed.

beawar (commented):
Please, follow the commit rules defined for the front-end projects (you can find them in the internal coding manifesto).
The issue code should go in the Refs field of the footer.

src/store/integrations/composer.tsx
frisonisland: Thanks, that's a nice suggestion. I tried to make as less changes as possible to avoid redefining or using a new component, not knowing the impact it would have on other components. I'll discuss it with my team

frisonisland: I reverted these changes and replaced the quicklink with link

AlessioCoser (approved):

FrancescoGottardi (approved):

beawar (approved):
Mine is a tech-review only

beawar:
@frisonisland Do we need to wait for the ux validation before merging this one?

frisonisland:
> @frisonisland Do we need to wait for the ux validation before merging this one?

No thanks, you can merge :slightly_smiling_face:

PR #434, opened 2024-06-10
refactor!: remove router inside boards

CataldoMazzilli:
BREAKING CHANGE: remove router inside boards
BREAKING CHANGE: change addBoard api
BREAKING CHANGE: change addBoardView api
BREAKING CHANGE: update Board type

Refs: SHELL-204

### DIFF
--- a/src/shell/boards/board.tsx
+++ b/src/shell/boards/board.tsx
@@ -4,16 +4,14 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-import React, { useEffect, useMemo } from 'react';
+import React, { useMemo } from 'react';
 
-import { createMemoryHistory } from 'history';
-import { find, startsWith } from 'lodash';
-import { Route, Router, useHistory } from 'react-router-dom';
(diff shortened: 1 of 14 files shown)


src/shell/boards/board.tsx
```diff
@@ -37,45 +35,19 @@ const BoardContainer = styled.div<{ show: boolean }>`
...
-							<view.component windowHistory={windowHistory} />
-						</BoardProvider>
-					</AppContextProvider>
-				</Route>
+				<AppContextProvider key={view.id} pkg={view.app}>
+					<BoardProvider id={board.id}>
+						<view.component />
+					</BoardProvider>
+				</AppContextProvider>
 			);
```
beawar: We could extract this component and just memoize the return value of the find. Then in the component return block we could conditionally render the AppContextProvider and the rest of it based on the memoized view. It seems cleaner to me.

beawar (approved):

beawar (approved):

rodleyorosa (approved):

PR #437, opened 2024-06-11
Mark zimbraPrefTimeZoneId preference setting as deprecated

AlessioCoser:
This is to avoid the wrong usage of the property.

We should not use it anymore because now we rely on the system timezone and that preference will be removed soon.

Refs: CO-1160

### DIFF
--- a/api-extractor/carbonio-shell-ui.api.md
+++ b/api-extractor/carbonio-shell-ui.api.md
@@ -149,7 +149,7 @@ export interface AccountSettingsPrefs {
     zimbraPrefOutOfOfficeSuppressExternalReply?: BooleanString;
     // (undocumented)
     zimbraPrefOutOfOfficeUntilDate?: GeneralizedTime;
-    // (undocumented)
+    // @deprecated
     zimbraPrefTimeZoneId?: string;
 }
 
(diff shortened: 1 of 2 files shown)


beawar:
Please add the Refs field as footer with the issue code.
You should also run the script build:local to update the api report, since the change involves the library (reason why the build is failing)

AlessioCoser:
> You should also run the script build:local to update the api report, since the change involves the library (reason why the build is failing)

👍🏼 done

AlessioCoser:
> Please add the Refs field as footer with the issue code. 

There aren't any issues that I'm aware of. While working on an unrelated task, we realized we were starting to use the timezone preference, which led us to discover that we should not use it. We then thought it would be useful to everyone to know that this preference is actually deprecated and the relative setting will be removed soon.

This is why we created that PR.

beawar:
Thanks for adding the ref. Notice that the build is still failing, because the deprecation mark requires a description for the suggestion of replacement

AlessioCoser:
@beawar now it should be ok. Can you merge it once the CI is completed?

Thanks.

beawar (approved):

rodleyorosa (approved):

CataldoMazzilli (approved):

PR #445, opened 2024-06-17
CO-1133 let user manage allowing and blocking addresses in Carbonio CE

mbarto:
:bell: needed by https://github.com/zextras/carbonio-mails-ui/pull/591

Adds saving amavis whitelist and blacklist senders from attrs.

:warning: we need to remove moment dependency before merging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
 				"i18next-http-backend": "^2.5.0",
 				"immer": "^10.0.3",
 				"lodash": "^4.17.21",
+				"moment": "^2.30.1",
 				"react": "^17.0.2",
 				"react-dom": "^17.0.2",
 				"react-i18next": "^12.3.1",
@@ -113,6 +114,9 @@
 					"optional": true
 				},
 				"lodash": {
+					"optional": true
+				},
+				"moment": {
 					"optional": true
 				},
 				"react": {
@@ -15373,6 +15377,14 @@
 			},
 			"engines": {
 				"node": ">=10"
+			}
+		},
+		"node_modules/moment": {
+			"version": "2.30.1",
+			"resolved": "https://registry.npmjs.org/moment/-/moment-2.30.1.tgz",
+			"integrity": "sha512-uEmtNhbDOrWPFS+hdjFCBfy9f2YoyzRpwcl+DqpC6taX21FzsTLQVbMV/W7PzNSX6x/bhC1zA3c2UQ5NzH6how==",
+			"engines": {
+				"node": "*"
 			}
 		},
 		"node_modules/mrmime": {
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
 		"i18next-http-backend": "^2.5.0",
 		"immer": "^10.0.3",
 		"lodash": "^4.17.21",
+		"moment": "^2.30.1",
 		"react": "^17.0.2",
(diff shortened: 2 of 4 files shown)


beawar (changes requested):
I am sorry, but I am totally against this change.

Edit settings function is deprecated from a while, because it breaks every rule of the clean code in our opinion. We are working on providing a way of notifying that the account has been updated, and that's why we have not removed it yet, but the modules should take care of sending their requests, without relying on this function, especially if the data are restricted to the module business logic. You added the section in mails settings, and this statement itself tell you that you are doing something related to mails. Shell should not know anything about mails business logic.

By placing the logic here, you are coupling on a business side shell and mails, creating a dependency between the two, as you also stated in the PR comment. And this is wrong in the idea of having a shell which is just the core that lets MFE work together.

We are trying to clean the shell by removing everything that does not concern it, and we made our goal pretty clear to every frontender in the last year. The proposed change is something that goes on the other direction, and I cannot accept it.

Also, why are you reintroducing moment? That does not make any sense, you can use moment in your modules, without shell having it.

mbarto:
Hi @beawar 

> Edit settings function is deprecated from a while, because it breaks every rule of the clean code in our opinion. We are working on providing a way of notifying that the account has been updated, and that's why we have not removed it yet, but the modules should take care of sending their requests, without relying on this function, especially if the data are restricted to the module business logic. You added the section in mails settings, and this statement itself tell you that you are doing something related to mails. Shell should not know anything about mails business logic.

Thanks for pointing that out, this is my first task on the UI modules, and I was not aware of this deprecation. I agree that it is something that should be handled by each module, so I will change the current solution to do that. Do you have examples on how we should do it or how other modules already do it?

> Also, why are you reintroducing moment? That does not make any sense, you can use moment in your modules, without shell having it.

This was just temporary so that I could test it on a VM, but I forgot to remove it before creating the PR.
If you agree I am going to close this PR right away and work to move the save logic inside mails-ui.

beawar:
> and I was not aware of this deprecation. 

We can work on making it more evident. We told it during one of our tribe meeting, and we added the jsdoc somewhere, but for sure we can also make it evident from the code.

> Do you have examples on how we should do it or how other modules already do it?

The general settings and the account settings their selves are not using the edit-settings anymore.
You can check the save method of either the general settings
https://github.com/zextras/carbonio-shell-ui/blob/e1421f66a9647a06f06df5a4b29886aefd1727bf/src/settings/general-settings.tsx#L99-L217

or the account settings

https://github.com/zextras/carbonio-shell-ui/blob/e1421f66a9647a06f06df5a4b29886aefd1727bf/src/settings/accounts-settings.tsx#L207-L295


> This was just temporary so that I could test it on a VM, but I forgot to remove it before creating the PR. If you agree I am going to close this PR right away and work to move the save logic inside mails-ui.

Even on the VM you don't need it, you should update the ui-sdk in mails project to have moment provided directly from the module.

I want to add that we have always been available to discuss technical solutions, especially when, and not restricted to, shell is involved, to avoid this kind of situation, where everyone time is kind of wasted (you spent time working on this, I spent time reviewing it, you now have to spend additional time re-doing it all).


You can close the PR if you don't have any further questions, or if you prefer to contact us directly.

mbarto:
We moved saving logic in the mails project, so we can close this one

PR #451, opened 2024-06-24
feat: expose a function to update settings store

gnekoz:
- Create and export `updateSettings` and `updateAccount` functions
- Refactor functions for merge props and prefs 

Refs: SHELL-226


src/lib.ts
```diff
@@ -211,9 +213,20 @@ export type {
...
+
+export type { EditSettingsBatchResponse } from './network/edit-settings';
+
+export type {
+	BatchRequest,
+	BatchResponse,
+	IdentityMods,
+	Mods,
+	PropsMods,
+	AttrsMods
```
beawar: Honestly, I don't think these types should be exported in the lib. These are types that are strictly related to the settings inside shell.

src/store/account/updaters.ts
```diff
@@ -0,0 +1,34 @@
...
+		...state,
+		account: {
+			...state.account,
+			displayName:
+				find(accountMods?.identity?.modifyList, (item) => item.id === state?.account?.id)?.prefs
+					.zimbraPrefIdentityName ?? state.account?.displayName,
+			identities: {
+				identity: updateIdentities(state, accountMods, response)
+			}
+		} as Account
```
beawar: Not a fan of the cast. It is hiding the fact that you are setting undefined in fields like displayName and identities.identity, which are declared as always-defined. Since the account might not be always set (shell allows un-authenticated executions), we should decide if we want to allow the update even if no account was previously set. If so, we should update the types accordingly, either setting all the needed fields as optional, or by setting some required field in the arg data.
Another possibility could be to prevent the update if no account was set before, and so the update should check that the state account is defined in order to update it, or it could throw if no account was set.

src/store/account/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+	ZimletProp
+} from '../../types/account';
+import type { Mods } from '../../types/network';
+
+export function mergePrefs(mods: Partial<Mods>, state: AccountState): AccountSettingsPrefs {
+	return reduce(
+		mods.prefs,
+		(acc, pref, key) => ({
+			...acc,
+			[key]: pref as string
```
beawar: The cast can be removed, it is not needed

src/types/account/index.ts
```diff
@@ -96,6 +98,12 @@ export type AccountSettingsAttrs = {
 	[key: string]: string | number | Array<string | number> | undefined;
 };
 
+export type UpdateSettings = (settingsMods: Partial<Mods>) => void;
+export type UpdateAccount = (
+	accountMods: Partial<Mods>,
+	response: Pick<EditSettingsBatchResponse, 'CreateIdentityResponse'>
+) => void;
```
beawar: I feel the two functions should not use the Mods or the EditSettingsBatchResponse util, I would prefer having them implementing their own types, so that we can 
- give them better namings, without having to change all the internal implementation
- avoid to expose the setting types, which were created in a sort of migration from js, and are all but clean

The types could also be place directly near the function definitions imo, since the types folder is full of duplicated types and is not quite organized, and we are trying to dismantle it

src/store/account/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+		state.settings.props
+	);
+}
+
+export function mergeAttrs(mods: Partial<Mods>, state: AccountState): AccountSettingsAttrs {
+	return reduce(
+		mods.attrs,
+		(acc, attr, key) => ({
+			...acc,
+			[key]: attr as string
```
beawar: the cast is not right, either the account setting attr should become unknown, or the mods one should become string.
We could probably decouple the mods arg from the Mods type, and pass only the attrs as argument, giving the function argument a smaller type.

In general, since the types are so mixed inside shell, I prefer having a bit of repetition in favor of the definition of better smaller types. I think that by doing so we can start isolating stuff.

src/network/edit-settings.ts
```diff
@@ -171,78 +172,16 @@ export const editSettings = (
...
-										...filter(
-											s.account.identities.identity,
-											(item) => !mods?.identity?.deleteList?.includes(item.id)
-										).filter((i) => i.name === 'DEFAULT')
-									]
-								)
-							: undefined
+					identity: updateIdentities(s, mods, r)
 				}
 			} as Account
```
beawar: We could replace the entire block with the updateSettings and updateAccount functions, couldn't we?

src/types/account/index.ts
```diff
@@ -96,6 +98,12 @@ export type AccountSettingsAttrs = {
 	[key: string]: string | number | Array<string | number> | undefined;
 };
 
+export type UpdateSettings = (settingsMods: Partial<Mods>) => void;
+export type UpdateAccount = (
+	accountMods: Partial<Mods>,
+	response: Pick<EditSettingsBatchResponse, 'CreateIdentityResponse'>
```
beawar: This arg is misleading in my opinion. It is called response, but it could be better to name it identities, and make it unrelated to the network response

src/lib.ts
mbarto: done

src/store/account/updaters.ts
mbarto: done

src/store/account/utils.ts
mbarto: done

src/types/account/index.ts
mbarto: Should be better now

src/store/account/utils.ts
mbarto: done

src/types/account/index.ts
mbarto: should be better now

src/network/edit-settings.ts
mbarto: We changed this by mistake, because we didn't notice it was deprecated code. Thanks for spotting that out. For this reason we reverted the change. Is it ok for you?

src/types/account/index.ts
```diff
@@ -96,6 +97,9 @@ export type AccountSettingsAttrs = {
 	[key: string]: string | number | Array<string | number> | undefined;
 };
 
+export type UpdateSettings = (settingsMods: Partial<Mods>) => void;
```
beawar: This is still using the Mods type, wouldn't it be better to use the AccountSettings? The Mods type includes additional fields which are not updated in the implementation you made. The AccountSettings is instead the exact representation of what will be updated.
Another solution could be creating a specific type for this function arg

beawar: I'll add an info. We should try to start using interfaces to declare types that could be extended (and remove the generic string key that is present right now only for compatibility with the previous types), so that modules can augment them to add fields strictly related to the module itself.
Specifically, if you need to extend the AccountSettingsPrefs, or the AccountSettingsAttrs, you can
1) refactor the type to be an interface if it is not already
2) export it if not already exported
3) augment the interface inside the module, in order to add the specific fields for that interface

A practical example
Let's say you are managing an account pref not listed inside the AccountSettingsPrefs of the shell.
You can have it type-checked by creating the module declaration for the shell, and extend in it the interface
```ts
import * as Shell from '@zextras/carbonio-shell-ui';

declare module '@zextras/carbonio-shell-ui' {
	interface AccountSettingsPrefs extends Shell.AccountSettingsPrefs {
		yourModulePref: 'value1' | 'value2';
	}
}
```

gnekoz: > I'll add an info. We should try to start using interfaces to declare types that could be extended (and remove the generic string key that is present right now only for compatibility with the previous types), so that modules can augment them to add fields strictly related to the module itself. Specifically, if you need to extend the AccountSettingsPrefs, or the AccountSettingsAttrs, you can
> 
> 1. refactor the type to be an interface if it is not already
> 2. export it if not already exported
> 3. augment the interface inside the module, in order to add the specific fields for that interface
> 
> A practical example Let's say you are managing an account pref not listed inside the AccountSettingsPrefs of the shell. You can have it type-checked by creating the module declaration for the shell, and extend in it the interface
> 
> ```ts
> import * as Shell from '@zextras/carbonio-shell-ui';
> 
> declare module '@zextras/carbonio-shell-ui' {
> 	interface AccountSettingsPrefs extends Shell.AccountSettingsPrefs {
> 		yourModulePref: 'value1' | 'value2';
> 	}
> }
> ```

We can talk about it but for now it's out of context considering the current task

gnekoz: > This is still using the Mods type, wouldn't it be better to use the AccountSettings? The Mods type includes additional fields which are not updated in the implementation you made. The AccountSettings is instead the exact representation of what will be updated. Another solution could be creating a specific type for this function arg

We know that the current type is not optimal, but changing it will have a bigger impact on the existing code.
Our approach has been a conservative one, re-using the existing types, to not change too much untested code with the risk of producing BCs.

nubsthead (dismissed):

src/types/network/index.ts
```diff
@@ -133,9 +134,12 @@ export type IdentityMods = {
 
 export type PrefsMods = Record<string, unknown> & AccountSettingsPrefs;
 
+export type AttrsMods = Record<string, unknown> & AccountSettingsAttrs;
+
 export interface Mods extends Record<string, Record<string, unknown> | undefined> {
 	props?: PropsMods;
 	prefs?: PrefsMods;
+	attrs?: AttrsMods;
```
beawar: The attrs field here is no more used and can be cleaned up

beawar (commented):
I think we are there. Just clean up the Mods type, since it has remained dirty from initial changes.

nubsthead (approved):

beawar (approved):

PR #452, opened 2024-06-25
feat: change message for overquota

rodleyorosa:
Refs: SHELL-200


src/settings/components/general-settings/user-quota.tsx
beawar, line 1: We decided that we would have used the standard message with the percentage for each case, aligning it to Files (check the comments of the issue)

src/settings/components/general-settings/user-quota.test.tsx
```diff
@@ -6,53 +6,80 @@
...
 describe('User Quota', () => {
-	describe('Quota description', () => {
-		it.each([
-			['even if it is 0', 0],
-			['if it is less than zimbraMailQuota', quotaMax - 1],
-			['if it is higher than zimbraMailQuota', quotaMax + 1]
-		])('should render the % of quota used message %s', (description, quotaUsed) => {
-			setupAccountStore(quotaUsed);
+	it.each([0, -1])(
+		'should show the string "[used] of unlimited space if zimbraMailQuota is %s',
```
beawar:
```suggestion
		'should show the string "[used] of unlimited space" if zimbraMailQuota is %s',
```

src/settings/components/utils.ts
```diff
@@ -1383,3 +1383,18 @@ export function calculateNewIdentitiesState(
 	filteredAndModified.splice(-1, 0, ...addedIdentities);
 	return filteredAndModified;
 }
+
+/**
+ * Format a size in byte as human-readable
+ */
+export const humanFileSize = (inputSize: number): string => {
```
beawar, line 1390: Copy the tests too

beawar (approved):

nubsthead (approved):

PR #462, opened 2024-07-08
feat: allow update of account signatures

dhavaldodiya:
Expose the updateAccount function to update the identities and signatures in account store.

Refs: SHELL-230 CO-1136


mbarto (dismissed):

src/store/account/utils.test.ts
```diff
@@ -317,4 +323,144 @@ describe('utils', () => {
...
+			if (!identityToDeleteId) {
+				throw new Error('Identity not found');
+			}
+
+			const mods: IdentityMods = {
+				deleteList: [identityToDeleteId]
+			};
+			expect(updateIdentities(state, mods, [])).toEqual(
+				expect.not.arrayContaining([expect.objectContaining({ id: identityToDeleteId })])
+			);
```
beawar: Why not using not.toContain instead of toEqual.not.arrayContaining? not.toContain seems more readable to me
```suggestion
			expect(updateIdentities(state, mods, [])).not.toContain(
				expect.objectContaining({ id: identityToDeleteId })
			);
```

src/store/account/utils.test.ts
```diff
@@ -317,4 +323,144 @@ describe('utils', () => {
...
+		it('should return the original identities and the new identities', () => {
+			setupAccountStore();
+			const state = useAccountStore.getState();
+			const newIdentity: Identity = {
+				_attrs: {},
+				id: faker.string.uuid()
+			};
+			expect(updateIdentities(state, {}, [newIdentity])).toEqual(
+				expect.arrayContaining([...(state.account?.identities.identity ?? []), newIdentity])
+			);
```
beawar: This is hiding the fact that the identities have an order.
It might be worth it to check the order too.

beawar (approved):

nubsthead (approved):

PR #464, opened 2024-07-22
feat: warn user about incoming expiration of session

beawar:
Show some snackbar to let know user the session is expiring soon. Once session is expired, redirect user to login page by performing a programmatic logout.

Refs: SHELL-207

### DIFF
--- a/src/boot/loader.tsx
+++ b/src/boot/loader.tsx
@@ -6,7 +6,7 @@
 
 import React, { useCallback, useEffect, useState } from 'react';
 
-import { Modal, Padding, Text } from '@zextras/carbonio-design-system';
+import { Modal, Padding, Text, useSnackbar } from '@zextras/carbonio-design-system';
 import { find } from 'lodash';
 import { useTranslation } from 'react-i18next';
 
@@ -16,6 +16,7 @@ import { IS_FOCUS_MODE } from '../constants';
 import { getComponents } from '../network/get-components';
 import { getInfo } from '../network/get-info';
 import { loginConfig } from '../network/login-config';
+import { logout } from '../network/logout';
 import { goToLogin } from '../network/utils';
 import { useAccountStore } from '../store/account';
 import { useAppStore } from '../store/app';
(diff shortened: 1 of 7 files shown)


gnekoz (dismissed):

src/boot/loader.tsx
```diff
@@ -105,6 +117,78 @@ export const Loader = (): React.JSX.Element => {
...
+						label: t(
+							'snackbar.expiration.threeMinutes',
+							"Your session will expire in 60 seconds. After that, you'll be redirected to the login page."
+						),
+						actionLabel: t('snackbar.expiration.action', 'Go to login page'),
+						onActionClick: logoutFn,
+						replace: true
+					});
+					expirationTimeouts.push(setTimeout(logoutFn, Math.min(oneMinute, sessionLifetime)));
+				}, sessionLifetime - oneMinute)
```
CataldoMazzilli: just to be safe...can we wrap in a Math.max(0, sessionLifetime - oneMinute)?

src/boot/loader.tsx
```diff
@@ -105,6 +117,78 @@ export const Loader = (): React.JSX.Element => {
 		return () => {
 			unloadAllApps();
 		};
-	}, []);
+	}, [getSessionInfo]);
+
+	useEffect(() => {
+		const expirationTimeouts: NodeJS.Timeout[] = [];
+		const logoutFn = (): void => {
```
CataldoMazzilli, line 124: can we use directly logout?

beawar: Using logout directly cause the sonarlint error typescript:S6544
SonarLint: Promise-returning function provided to property where a void return was expected.

That's why I wrapped it in a non-promise-returning function

CataldoMazzilli (dismissed):

CataldoMazzilli (dismissed):

src/boot/loader.tsx
```diff
@@ -105,6 +117,81 @@ export const Loader = (): React.JSX.Element => {
...
+			const oneMinute = 60 * 1000;
+			expirationTimeouts.push(
+				setTimeout(
+					() => {
+						createSnackbar({
+							severity: 'warning',
+							key: 'one-minute-from-expiration-snackbar',
+							autoHideTimeout: Math.min(oneMinute, sessionLifetime),
+							label: t(
+								'snackbar.expiration.threeMinutes',
```
beawar:
```suggestion
								'snackbar.expiration.oneMinute',
```

CataldoMazzilli (approved):

rodleyorosa (approved):

CerDennis (approved):

PR #482, opened 2024-08-09
feat: show dynamic counter in 60 seconds snackbar session expiration

beawar:
Refs: SHELL-232

---

Translation changes:

- snackbar.expiration.oneMinute
defaultValue_one: "Your session will expire in {{count}} second. After that, you'll be redirected to the login page."
defaultValue_other: "Your session will expire in {{count}} seconds. After that, you'll be redirected to the login page."


src/boot/loader.tsx
```diff
@@ -64,6 +64,39 @@ export const LoaderFailureModal = ({
 	);
 };
 
+const ExpiringSessionDynamicLabel = ({
+	sessionLifetime
+}: {
+	sessionLifetime: number;
+}): React.JSX.Element => {
+	const oneMinute = 60 * 1000;
+	const initialCounter = Math.ceil(Math.min(sessionLifetime, oneMinute) / 1000);
```
CataldoMazzilli: Can we extract rows 72 and 73 and wrap into a function invoked inside state initial state?

CataldoMazzilli (dismissed):

src/boot/loader.tsx
```diff
@@ -64,6 +64,39 @@ export const LoaderFailureModal = ({
...
+
+	useEffect(() => {
+		const interval = setInterval(() => {
+			setCount((prevState) => prevState - 1);
+		}, 1000);
+
+		return (): void => {
+			clearInterval(interval);
+		};
+	}, [initialCounter]);
```
beawar: This should not be a dependency

src/boot/loader.tsx
beawar: Yes, I'm on it

CataldoMazzilli (approved):

rodleyorosa (approved):

PR #483, opened 2024-08-14
Shell 219 enhance primary bar component to include common customizations

CataldoMazzilli:


src/shell/badge-wrap.tsx
```diff
@@ -26,20 +26,37 @@ const MiniBadge = styled(Badge)`
...
 			<Container width={'3rem'} height={'3rem'} style={{ position: 'relative' }} ref={ref}>
-				{badge.show && (
-					<MiniBadge
-						color={'gray6'}
-						backgroundColor={badge.color ?? 'primary'}
-						data-testid={'badge-counter'}
-						value={badge.showCount ? badge.count ?? 0 : ''}
-					/>
-				)}
+				{(badge.show && badge.icon && <MiniIcon icon={badge.icon} color={badge.color} />) ||
```
beawar, line 46: The new badge supports the icon inside it, I feel it would be better to use that instead of rendering an icon

eugeniamuff: @beawar Together with @CMazzilli  we tried to use the badge but we saw that its icon is too small. For this reason we thought it would be better to use the icon as we currently use on Files. 

We will re-evaluate the use of the badge when we begin the following issue: https://zextras.atlassian.net/browse/CDS-241

beawar (commented):
There is a missing point to do: the PrimaryBarElement should pass down all the props to the custom component, instead of only the active prop

beawar (approved):

rodleyorosa (approved):

PR #486, opened 2024-08-19
fix: noOp endless loop

mgarbin:
Refs: SHELL-239


src/store/network/utils.ts
```diff
@@ -47,11 +47,13 @@ export const parsePollingInterval = (settings: AccountSettings): number => {
...
-export const getPollingInterval = (res: RawSoapResponse<{ waitDisallowed?: number }>): number => {
+export const getPollingInterval = (res: RawSoapResponse<{
+	NoOpResponse?: { waitDisallowed? : boolean }
+}>): number => {
 	const { pollingInterval } = useNetworkStore.getState();
 	const { settings } = useAccountStore.getState();
-	const waitDisallowed = res.Body && !('Fault' in res.Body) && res.Body.waitDisallowed;
-	const fault = res.Body && 'Fault' in res.Body && res.Body.Fault;
+	const waitDisallowed = res?.Body && !('Fault' in res?.Body) && res?.Body?.NoOpResponse && res.Body?.NoOpResponse?.waitDisallowed;
+	const fault = res?.Body && 'Fault' in res?.Body && res?.Body?.Fault;
```
beawar: - res cannot be undefined by definition.
- All these "?" are useless. The optional chain guarantee that the object is not undefined after the first check.
- The condition can be simplified to make it more readable.

```suggestion
	const waitDisallowed =
		res.Body && !('Fault' in res.Body) && res.Body.NoOpResponse?.waitDisallowed;
	const fault = res.Body && 'Fault' in res.Body && res.Body.Fault;
```

Also, eslint is failing because the code is not formatted under the current rules.

beawar (commented):
Build is still failing

beawar (approved):

CataldoMazzilli (approved):

CataldoMazzilli:
I approve but this part needs a deeper refactor

beawar:
> I approve but this part needs a deeper refactor

I agree, by just implementing a couple of tests multiple problems on the parsing are emerged.
We need to review the polling entirely (consider also that it is related to a mail pref, set by mails, I honestly think this should not be shell code)

PR #493, opened 2024-08-29
fix: hide privacy and remove account subsections from settings

beawar:
Hide privacy entry inside secondary bar following the same logic of the settings component.

Remove accounts subsection since it is not used anymore

Refs: SHELL-241

### DIFF
--- a/src/boot/app/default-views.test.tsx
+++ b/src/boot/app/default-views.test.tsx
@@ -1,0 +1,166 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Zextras <https://www.zextras.com>
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import React from 'react';
+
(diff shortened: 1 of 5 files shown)


src/boot/app/default-views.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+		useAccountStore.setState(
+			produce((state) => {
+				state.settings.attrs.zimbraFeatureOptionsEnabled = 'FALSE';
+			})
+		);
+		setup(<DefaultViewsRegister />);
+		expect(Object.keys(useAppStore.getState().routes)).not.toContain(SETTINGS_APP_ID);
+	});
+
+	it('should test', () => {
```
rodleyorosa: is this test necessary?

CataldoMazzilli (approved):

rodleyorosa (approved):

PR #510, opened 2024-09-23
fix: slow response in email composer

mananpatel13589:
Ref : [CO-1454](https://zextras.atlassian.net/browse/CO-1454)

[CO-1454]: https://zextras.atlassian.net/browse/CO-1454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/store/integrations/composer.tsx
+++ b/src/store/integrations/composer.tsx
@@ -78,11 +78,10 @@ const Composer: FC<ComposerProps> = ({
 	...rest
 }) => {
 	const _onEditorChange = useCallback<NonNullable<EditorProps['onEditorChange']>>(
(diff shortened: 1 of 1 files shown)


beawar (changes requested):
We tried your solution with the email attached in the issue.
Unfortunately, we cannot see the performance improvement with these changes.
By seeing the documentation and the code, we think that a possible improvement could be reached by using the uncontrolled mode.

cc @CMazzilli

beawar:
I'm closing this since it's stalled

PR #513, opened 2024-09-27
fix: redirect to login even if logout requests fail

beawar:
Refs: SHELL-248


CataldoMazzilli (approved):

src/network/logout.test.ts
```diff
@@ -0,0 +1,156 @@
...
+			useLoginConfigStore.setState({ carbonioWebUiLogoutURL: 'custom logout' });
+			controlConsoleError('Failed to fetch');
+			const goToFn = jest.spyOn(utils, 'goTo').mockImplementation();
+			server.use(http.get('/logout', () => HttpResponse.error()));
+			await logout();
+			await jest.advanceTimersToNextTimerAsync();
+			expect(goToFn).toHaveBeenCalled();
+		});
+
+		it('should redirect to login page if EndSession request succeeded with Fault', async () => {
```
rodleyorosa, line 124: there are two `should redirect to login page if EndSession request succeeded with Fault` tests in this file

rodleyorosa: nvm, they are in different `describe`

rodleyorosa (approved):

PR #514, opened 2024-09-27
fix: avoid loading apps if getInfo fails

beawar:
Refs: SHELL-249


beawar:
It requires further discussion

PR #522, opened 2024-10-07
feat!: remove deprecated constants

rodleyorosa:
Refs: SHELL-158


beawar:
To reopen after removal of folders and tags

PR #531, opened 2024-10-24
fix: fix parse of polling interval to consider all duration unit

beawar:
Refs: SHELL-240

### DIFF
--- a/src/network/fetch.test.ts
+++ b/src/network/fetch.test.ts
@@ -96,7 +96,7 @@ describe('Fetch', () => {
 				)
 			);
 
-			expect(useNetworkStore.getState().noOpTimeout).not.toBeDefined();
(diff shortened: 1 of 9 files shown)


src/network/fetch.test.ts
```diff
@@ -74,4 +80,152 @@ describe('Fetch', () => {
...
+						HttpResponse.json({
+							Body: {},
+							Header: {
+								context: {}
+							}
+						})
+				)
+			);
+
+			expect(useNetworkStore.getState().noOpTimeout).not.toBeDefined();
```
rodleyorosa:
```suggestion
			expect(useNetworkStore.getState().noOpTimeout).toBeUndefined();
```

CataldoMazzilli (approved):

rodleyorosa (approved):

PR #532, opened 2024-10-30
Export expand boards hook

giuliano176:
ref: CO-1077


beawar (approved):
Not sure whether this is a chore, since it is changing the API. A feat maybe?

CataldoMazzilli (approved):

PR #542, opened 2024-11-08
fix: include zimbraAllowFromAddress in sender address list

aheeva-yuliya:
This PR implements the functionality to retrieve values from the `zimbraAllowFromAddress` setting and add them to the existing address list that a user can use as their "sender" address.

Refs: [CO-1364](https://zextras.atlassian.net/browse/CO-1364)

[CO-1364]: https://zextras.atlassian.net/browse/CO-1364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/settings/components/utils.ts
```diff
@@ -496,13 +496,32 @@ export function defaultAsFirstOrderIdentities(identities: Array<Identity>): Arra
...
+ * Return the requested attribute value as an array if present, or an empty array otherwise.
+ *
+ * @param settings
+ * @param key
+ * @returns
+ */
+export function getAttributeIfPresent(settings: AccountSettings, key: string): Array<string> {
+	const attributeValue = settings.attrs[key];
+	if (attributeValue) {
+		if (isArray(attributeValue)) {
```
beawar: prefer `Array.isArray` instead of lodash (we are trying to remove lodash, so in almost every refactor we try to replace it with pure js utils)

src/settings/components/utils.ts
```diff
@@ -496,13 +496,32 @@ export function defaultAsFirstOrderIdentities(identities: Array<Identity>): Arra
...
 }
 
+/**
+ * Return the requested attribute value as an array if present, or an empty array otherwise.
+ *
+ * @param settings
+ * @param key
+ * @returns
+ */
+export function getAttributeIfPresent(settings: AccountSettings, key: string): Array<string> {
```
beawar: the type of key should be `keyof AccountSetting['attrs']`

src/settings/components/utils.ts
```diff
@@ -496,13 +496,32 @@ export function defaultAsFirstOrderIdentities(identities: Array<Identity>): Arra
...
+	const attributeValue = settings.attrs[key];
+	if (attributeValue) {
+		if (isArray(attributeValue)) {
+			return [...(attributeValue as string[])];
```
beawar: there's no need to spread attributeValue to create the array, since it is already.

src/settings/components/utils.ts
beawar: TBH I don't think this util should always return strings. It should return the proper type of the attribute. This can be achieved through generics, and with a proper update of the type AccountSettingsAttrs

src/settings/components/utils.ts
```diff
@@ -533,15 +552,8 @@ export const getAvailableEmailAddresses = (
...
-	if (settings.attrs.zimbraMailAlias) {
-		if (isArray(settings.attrs.zimbraMailAlias)) {
-			result.push(...(settings.attrs.zimbraMailAlias as string[]));
-		} else {
-			result.push(String(settings.attrs.zimbraMailAlias));
-		}
-	}
-
+	result.push(...getAttributeIfPresent(settings, 'zimbraMailAlias'));
+	result.push(...getAttributeIfPresent(settings, 'zimbraAllowFromAddress'));
```
beawar: The 2 fields zimbraMailAlias and zimbraAllowFromAddress need to be added to AccountSettingsAttrs with their proper value type

src/settings/components/utils.test.ts
```diff
@@ -103,3 +105,69 @@ describe('humanFileSize function', () => {
...
+			attrs: {},
+			prefs: {},
+			props: [{ name: 'zimlet1', zimlet: 'propValue1', _content: 'contentValue1' }]
+		};
+
+		const result = getAttributeIfPresent(settings, 'key');
+		expect(result).toEqual([]);
+	});
+
+	it('should call _.isArray with the correct argument', () => {
```
beawar: What is the goal of this test? It's testing the implementation instead of the input/output

src/settings/components/utils.ts
beawar: The assertion is wrong. If the attribute is an array of number, the returned value will be an array of number, while the type declare it will always be an array of strings

src/settings/components/utils.ts
```diff
@@ -496,13 +496,39 @@ export function defaultAsFirstOrderIdentities(identities: Array<Identity>): Arra
...
+		if (Array.isArray(attributeValue)) {
+			return attributeValue as AccountSettings['attrs'][T] extends (string | number)[]
+				? (string | number)[]
+				: never;
+		}
+		return [attributeValue] as (string | number)[];
+	}
+	return [] as (string | number)[];
+}
+
```
beawar: I'm wondering if it is really needed something like this.
We could have a much simpler function if we move out the reading of the attribute. The util could focus only on returning an array starting from a value that could be both an array or not.
Something like this
```ts
export function asArray<T>(value: T | T[] | undefined): T[] {
	if (value !== undefined) {
		if (Array.isArray(value)) {
			return value;
		}
		return [value];
	}
	return [];
}
```

And then it could be used like this
```ts
result.push(...asArray(settings.attrs.zimbraMailAlias));
result.push(...asArray(settings.attrs.zimbraAllowFromAddress));
```

src/types/account/index.ts
```diff
@@ -102,6 +102,8 @@ export type AccountSettingsAttrs = {
 	zimbraFeatureOptionsEnabled?: BooleanString;
 	zimbraIdentityMaxNumEntries?: number;
 	[key: string]: string | number | Array<string | number> | undefined;
+	zimbraMailAlias?: string | Array<string>;
+	zimbraAllowFromAddress?: string | Array<string>;
```
beawar, line 105: Can you move them above the generic [key: string]?

src/settings/components/utils.test.ts
```diff
@@ -103,3 +103,59 @@ describe('humanFileSize function', () => {
 		expect(() => humanFileSize(1024 ** 9, undefined)).toThrow('Unsupported inputSize');
 	});
 });
+
+describe('asArray', () => {
```
CataldoMazzilli, line 107: asArray is agnostic of settings context so now these tests are out of context

src/settings/components/utils.ts
```diff
@@ -496,13 +496,31 @@ export function defaultAsFirstOrderIdentities(identities: Array<Identity>): Arra
 	return result;
 }
 
+/**
+ * Return the requested attribute value(s) as an array if present, or an empty array otherwise.
+ *
+ * @param settings - The account settings containing the attributes.
+ * @param key - The key of the attribute in the `attrs` object.
+ * @returns An array containing the attribute value(s) (string | number).
+ */
```
CataldoMazzilli, line 512: this doc is outdated

src/settings/components/utils.ts
```diff
@@ -533,14 +551,8 @@ export const getAvailableEmailAddresses = (
...
-	// Adds all the aliases
-	if (settings.attrs.zimbraMailAlias) {
-		if (isArray(settings.attrs.zimbraMailAlias)) {
-			result.push(...(settings.attrs.zimbraMailAlias as string[]));
-		} else {
-			result.push(String(settings.attrs.zimbraMailAlias));
-		}
-	}
+	result.push(...asArray(settings.attrs.zimbraMailAlias));
+	result.push(...asArray(settings.attrs.zimbraAllowFromAddress));
```
CataldoMazzilli: result.push(...asArray(settings.attrs.zimbraMailAlias), ...asArray(settings.attrs.zimbraAllowFromAddress));

src/settings/components/utils.test.ts
```diff
@@ -103,3 +103,29 @@ describe('humanFileSize function', () => {
...
+	it('should return an array when the value is an array', () => {
+		const result = asArray(['value1', 'value2']);
+		expect(result).toEqual(['value1', 'value2']);
+		expect(Array.isArray(result)).toBe(true);
```
beawar: All these expect are redundant since the fact that result is an array is already checked by the comparison with the `[...]`

beawar (commented):
Tests can be cleaned up a bit more, everything else looks good to me now

beawar (approved):
LGTM

CataldoMazzilli (approved):

zextrasGiovanniDefacci (approved):

PR #543, opened 2024-11-12
feat: added Hungarian language

FrancescoGottardi:
- added support for Hungarian language


beawar (commented):
This is a feat, not a chore, since it is adding something.

Have you checked date-fns and tinymce locales to see if the key match?

FrancescoGottardi:
> This is a feat, not a chore, since it is adding something.
> 
> Have you checked date-fns and tinymce locales to see if the key match?



> This is a feat, not a chore, since it is adding something.
> 
> Have you checked date-fns and tinymce locales to see if the key match?

Hi,
i added the one related to the tinymce, the dats fns instead was fine;
- https://github.com/date-fns/date-fns/tree/main/src/locale/hu
- https://github.com/mklkj/tinymce-i18n/blob/master/langs/hu_HU.js

beawar:
> i added the one related to the tinymce
> 
> * https://github.com/mklkj/tinymce-i18n/blob/master/langs/hu_HU.js

I'm sorry, but this is not the official language pack for the version we are using (v6). See the documentation here https://www.tiny.cloud/docs/tinymce/6/ui-localization/ (the link is present also in the LocaleDescriptor type)

Also, if you try the UI and set the UI language to Hungarian, you can see the composer is not translated. And that's because the language pack is not available.
You need to also add the language file inside the assets/tinymce/langs folder, which is used by the Composer component to retrieve the translations

assets/tinymce/langs/carbonio-head.svg
beawar, line 1: Wrong move

beawar (approved):
Looks good now

keshavbhatt (approved):

CataldoMazzilli (approved):

beawar:
@Gammax92 @MatteoPerdon Remember to add strings in weblate

PR #548, opened 2024-11-21
feat: expose upsertApp to add or update display label

rodleyorosa:
Refs: SHELL-256

### DIFF
--- a/src/shell/boards/board-container.tsx
+++ b/src/shell/boards/board-container.tsx
@@ -35,7 +35,7 @@ import {
 	LOCAL_STORAGE_BOARD_SIZE,
 	PRIMARY_BAR_WIDTH
 } from '../../constants';
-import { getApp } from '../../store/app';
+import { useAppStore } from '../../store/app';
 import {
 	closeAllBoards,
 	closeBoard,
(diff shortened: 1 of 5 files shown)


beawar (approved):

src/shell/boards/board-container.tsx
```diff
@@ -210,13 +211,13 @@ export const BoardContainer = (): React.JSX.Element | null => {
 							label={boards[boardId].title}
 							icon={boards[boardId].icon}
 							selected={boardId === current}
-							app={getApp(boards[boardId].app)()?.display}
+							app={apps[boards[boardId].app].display}
```
CataldoMazzilli, line 214: apps[boards[boardId].app]   could it be undefined?

rodleyorosa: actually `apps` is always set on first render by the `getComponents()` function (https://github.com/zextras/carbonio-shell-ui/blob/devel/src/network/get-components.ts) and `boards` is set from the other modules.

In this case, if a module does not set the `upsertApp` function exported by Shell, the string will not be simply translated.

Do you think there is a corner case to handle?

CataldoMazzilli (approved):

PR #556, opened 2024-12-06
feat!: separate search module

beawar:
- Remove search module from shell and create a separate module zextras/carbonio-search-ui.
- Add shell as fallback for translations. The order of languages should now be module/{lang}.json > module/en.json > shell/{lang}.json > shell/en.json
- Allow to set generic to useIntegratedFunction, useIntegratedComponent, getIntegratedFunction and getIntegratedComponent to better type the returned integration.

BREAKING CHANGE: remove all exports related to the search. Search functions and components must be used through the integrations exposed by the search module.
Refs: SHELL-260


src/store/i18n/store.ts
```diff
@@ -94,17 +95,26 @@ export const useI18nStore = create<I18nState & I18nActions>()((set) => ({
...
 							// for all options read: https://www.i18next.com/overview/configuration-options
 							.init({
 								...defaultI18nInitOptions,
 								lng: locale,
 								backend: {
-									loadPath:
-										app.name === SHELL_APP_ID
-											? `${BASE_PATH}/i18n/{{lng}}.json`
-											: `${dropRight(app.js_entrypoint.split('/')).join('/')}/i18n/{{lng}}.json`
+									backends: [HttpBackend, HttpBackend],
```
CataldoMazzilli, line 105: 2 times HttpBackend?

CataldoMazzilli (dismissed):

src/store/i18n/store.ts
beawar: Yes, the second one is the fallback which will use the second configuration of the backendOptions field

beawar: Yes, the second one is the fallback which will use the second configuration of the backendOptions field

rodleyorosa (dismissed):

CataldoMazzilli (dismissed):

rodleyorosa (dismissed):

CataldoMazzilli (dismissed):

CataldoMazzilli (approved):

rodleyorosa (approved):

PR #560, opened 2024-12-18
feat: export theme augmentation

beawar:
Remove augmentation of styled-components in favor of the
carbonio-design-system one.
Export theme augmentation inside lib so that it is available to modules.

* chore: update thirdparties


beawar:
- [x] check if the ds augmentation is considered during the build of the lib
- [x] check if the ds augmentation can be included in the lib build, so that it is available for modules by installing shell dependency

CataldoMazzilli (approved):

lisaparma (approved):

rodleyorosa (approved):

PR #621, opened 2025-06-18
Co 1990 auto reply message should give indication that the max characters limit has exceeded

nubsthead:
refs: CO-1990

To merge after: https://github.com/zextras/carbonio-design-system/pull/511


src/settings/components/settings-header.tsx
```diff
@@ -34,13 +34,15 @@ export type SettingsHeaderProps = {
...
+	hasError?: boolean;
 };
 
 export const SettingsHeader = ({
 	onSave,
 	onCancel,
 	isDirty,
-	title
+	title,
+	hasError = false
```
rodleyorosa, line 45: is the default to false necessary?

nubsthead: Not really, should I take it off?

rodleyorosa: i would say yes, it seems to be a useless default

src/settings/general-settings.tsx
```diff
@@ -222,7 +226,13 @@ const GeneralSettings = (): React.JSX.Element => {
 
 	return (
 		<>
-			<SettingsHeader title={title} onCancel={onCancel} onSave={onSave} isDirty={isDirty} />
+			<SettingsHeader
+				title={title}
+				onCancel={onCancel}
+				onSave={onSave}
+				isDirty={isDirty}
+				hasError={hasError}
```
rodleyorosa, line 233: use directly `outOfOfficeError` instead of creating a memo `hasError`
```suggestion
				hasError={outOfOfficeError}
```

rodleyorosa (dismissed):

src/settings/general-settings.tsx
nubsthead: I think it is better to keep the two logic split. There could be more errors to handle and they can all handled inside that memo

src/settings/general-settings.test.tsx
```diff
@@ -23,6 +27,34 @@ describe('General setting', () => {
 	const { defaultI18n } = useI18nStore.getState();
 	const localeArray = localeList(defaultI18n.t);
 
+	test('When there are changes and an error, the discard button is enabled but save button is disabled', async () => {
```
CataldoMazzilli, line 30: the error come from the stored state of zimbraPrefOutOfOfficeReply and not from a user interaction: you are simulating that the wrong value is retrieved from the BE. Is it as intended?

src/settings/components/general-settings/out-of-office-settings.tsx
```diff
@@ -247,56 +266,126 @@ export const OutOfOfficeSettings = ({
...
 			<FormSubSection>
-				<Container gap={'0.5rem'}>
-					<Select
-						items={prefOutOfOfficeReplyEnabledSelectItems}
-						label={t('label.out_of_office', 'Out of Office')}
+				<Container gap={'0.5rem'} crossAlignment={'flex-start'}>
+					<Switch
+						value={prefOutOfOfficeReplyEnabled}
+						label={t('label.send_auto_reply', 'Send auto-reply')}
+						onClick={prefOutOfOfficeReplyEnabledOnClick}
```
CataldoMazzilli, line 349: Why we need to use both onClick and onChange?

src/settings/components/general-settings/out-of-office-settings.tsx
```diff
@@ -23,33 +25,15 @@ import { type TFunction } from 'i18next';
...
-
-export const buildItemsPrefOutOfOfficeReplyEnabled = (
-	t: TFunction
-): Array<
-	SelectItem<
-		NonNullable<CoercedPrefType<AccountSettingsPrefs['zimbraPrefOutOfOfficeReplyEnabled']>>
-	>
-> => [
-	{
-		label: t('settings.out_of_office.send_auto_replies', 'Send auto-replies'),
```
CataldoMazzilli, line 44: Are new keys pushed and no-more used removed?

src/settings/general-settings.test.tsx
nubsthead: In this test I am simulating any error, it does not matter what causes it. If you want I can add one more test to simulate a specific error from a user interaction

src/settings/components/general-settings/out-of-office-settings.tsx
nubsthead: I used both because in the onCLick action we don't receive the actual value setted by the click to the Switch component. We only receive a click event that makes it almost impossible to manage logic which relies on the current value.

On The other hand the onChange is receiving the new setted value of the Switch component.

So I am using the onClick to update the value and the onChange to manage the logic depending from that value.

I believe the onClick should receive the setted value instead of a click event in order to be able to use only one of the 2 functions instead of beeing forced to use both.
Also it is the first time I use this component so I am probably using it improperly, let me know if you have any suggestion on how to use it.

CataldoMazzilli (approved):

rodleyorosa (approved):

lisaparma (approved):

PR #868, opened 2026-07-15
chore(deps-dev): weekly devDependencies update 2026-07-15

CataldoMazzilli:
## Weekly devDependencies update — 2026-07-15

Automated bundle of all **devDependencies** bumped to the latest minor/patch within the same major.

### Updates

| Package | Old | New | Minor jump |
|---------|-----|-----|------------|
| @semantic-release/github | 12.0.8 | 12.0.9 | 0 |
| @types/node | 22.19.19 | 22.20.1 | +1 |
| @vitest/coverage-istanbul | 4.1.9 | 4.1.10 | 0 |
| @zextras/carbonio-ui-sdk | 2.3.6 | 2.3.12 | 0 |
| autoprefixer | 10.5.2 | 10.5.3 | 0 |
| msw | 2.14.6 | 2.15.0 | +1 |
| semantic-release | 25.0.5 | 25.0.7 | 0 |
| vitest | 4.1.9 | 4.1.10 | 0 |
| webpack-dev-server | 5.2.5 | 5.2.6 | 0 |

### Notes

- Base branch: `devel`
- Command: `ncu -u --target minor --dep dev && pnpm install`
- Only `package.json` and `pnpm-lock.yaml` were modified.
- **`webpack` (5.107.2 → 5.108.4) was split out into #870.** Its build failed inside this bundle (`sdk build` crashed with `TypeError: The 'compilation' argument must be an instance of Compilation`) due to duplicate webpack instances caused by a `postcss` peer mismatch between the project and `@zextras/carbonio-ui-sdk`. That bump needs pnpm `overrides` to deduplicate, so it gets its own focused review in #870. This bundle keeps the remaining 9 clean bumps and now builds green.

🤖 Generated by the `update-deps` skill.

PR #870, opened 2026-07-15
chore(deps-dev): update webpack to 5.108.4 with instance-safe DefinePlugin

CataldoMazzilli:
## Weekly dependency update — webpack (split from #868)

This PR was split out of the bundled devDependencies PR #868 because the `webpack` bump surfaced a real (latent) bug in our build setup that deserves a focused review.

### The bump

| Package | Old | New | Jump |
|---|---|---|---|
| `webpack` | `5.107.2` | `5.108.4` | +1 |

### What broke

webpack `5.108.4` tightens the `instanceof Compilation` check in `DefinePlugin.getCompilationHooks`. That turned a **pre-existing dual-webpack situation into a hard build failure**:

- `sdk build` compiles with the **sdk's own webpack copy** (the sdk resolves its toolchain internally via `require.resolve`).
- `carbonio.webpack.ts` imported `DefinePlugin` from the **project's webpack copy**.
- The plugin from one instance received a `Compilation` created by the other → `TypeError: The 'compilation' argument must be an instance of Compilation`.

It had always "worked" only because older webpack never checked instance identity on that path.

### The fix: instance-safe DefinePlugin (`compiler.webpack`)

`carbonio.webpack.ts` now instantiates DefinePlugin through **`compiler.webpack`** — webpack's official escape hatch (since 5.1) that hands plugins the classes of the instance actually running the build:

```ts
{
	apply(compiler: Compiler): void {
		new compiler.webpack.DefinePlugin({ ... }).apply(compiler);
	}
}
```

This is the same pattern `copy-webpack-plugin` / `html-webpack-plugin` already use internally (which is why they never crashed). The project's webpack copy is now **types-only** (`import type { Compiler }`): it never executes, so mixed instances are structurally impossible — with the sdk's current webpack (5.105.4 on this branch), with 5.108.3 after #868, and with any future version.

### Verified

- `pnpm run build` green locally **with the crash-condition pair in place** (project 5.108.4 + sdk 5.105.4).
- DefinePlugin injection confirmed in the output bundles: `COMMIT_ID`/`BASE_PATH`/`POSTHOG_API_*` are replaced at compile time, no bare identifiers left in `dist/*.js`.
- Full hook suite passed: type-check, lint, reuse, 716 tests, build + api-extractor.

### Review notes

- An earlier revision of this PR fixed the same problem with pnpm `overrides` (`webpack` + `postcss`) forcing a single deduplicated instance. It worked, but it treated the symptom and had to be replicated in every repo that uses the sdk. This revision fixes the cause in our own config instead and drops the overrides entirely.
- The same pattern is portable to the other Carbonio UI projects (files-ui, tasks-ui, …), which will hit the same crash whenever their webpack reaches 5.108.x.
- Long-term, the sdk could declare webpack/postcss as peerDependencies so a single instance is guaranteed by construction; with this fix that becomes a nice-to-have rather than a blocker.

---
🤖 Opened by the automated `update-deps` skill (weekly run 2026-07-15); reworked to the instance-safe fix on request.


CataldoMazzilli:
♻️ **Reworked.** Dropped the pnpm `overrides` (webpack/postcss) approach and fixed the root cause instead: `carbonio.webpack.ts` now creates DefinePlugin via `compiler.webpack` (the instance actually compiling), so the project's webpack copy is types-only and can never mix with the sdk's at runtime. Verified locally with the crash-condition pair (project 5.108.4 + sdk 5.105.4): build green, DefinePlugin constants correctly injected in the bundles. Body updated with details.

rodleyorosa (approved):

lisaparma (approved):

PR #907, opened 2026-09-10
chore(deps): update vitest to 5.0.0

CataldoMazzilli:
## What

Updates vitest to 5.0.0, together with its coverage provider — `@vitest/coverage-istanbul` pins vitest as an **exact** peer, so the two always move together.

| Package | Old | New |
|---|---|---|
| `vitest` | `4.1.11` | `5.0.0` |
| `@vitest/coverage-istanbul` | `4.1.10` | `5.0.0` |

This also realigns the pair: #905 bumped only `vitest` to 4.1.11 for the security advisory, leaving the provider behind on 4.1.10 against an exact `vitest: 4.1.10` peer.

## What vitest 5 required, and what it didn't

| Breaking change | Impact here |
|---|---|
| Requires Node.js 22 and Vite 6.4 | None — CI runs Node 22.23.1, Vite resolves to 7.3.2 |
| **Clear mocks by default before each test** | None — `clearMocks: true` was already explicit in the config (now redundant, kept for clarity) |
| **junit/json reporter output moves to `.vitest/`** | None — `outputFile: { junit: './junit.xml' }` is explicit, so the report stays where the pipeline reads it. Verified: `junit.xml` and `coverage/lcov.info` land in the same paths as before |
| Remove `sequential` in favour of `concurrent` | Not used |
| Fail the test when an async assertion is not awaited | No `.resolves` / `.rejects` / `expect.poll` in the suite |
| Remove deprecated entry points | Only `from 'vitest'` is imported |
| Config lookup no longer walks ancestor directories | Config is at the project root |
| `loupe.inspect` → pretty-format | Error formatting only |
| Updated `@sinonjs/fake-timers` | The global setup passes an explicit `toFake` list, so no default shifted under it |
| **`Assertion` type changed, `jest` namespace augmentations dropped** | **487 type errors — see below** |
| **Coverage globs no longer matched too eagerly** | **One more file measured — see below** |

`vitest-fail-on-console` (0.10.1) stays compatible: it peers on `@vitest/utils`, which vitest 5 still publishes.

## The type-level breaking change

Vitest 5 inlined the `expect` package. Two consequences:

`Assertion` went from one type parameter to two, the first being the return type that powers the new promise-aware matcher typing:

```ts
// vitest 4
interface Assertion<T> { … }

// vitest 5
interface Assertion<R extends void | Promise<void> = void, T = unknown>
  extends VitestAssertion<Chai.Assertion, R, T>, JestAssertion<R, T>, ChaiMockAssertion<R, T>, Matchers<R, T>
```

And more consequentially, matcher augmentations declared against the **global `jest` namespace** are no longer picked up. That is still how both matcher libraries in this project register:

```ts
// @testing-library/jest-dom 7.0.1 — types/jest.d.ts
declare global {
  namespace jest {
    interface Matchers<R = void, T = {}> extends TestingLibraryMatchers<…> {}
  }
}

// @emotion/jest 11.14.2 — types/index.d.ts
declare global {
  namespace jest {
    interface Matchers<R, T> {
      toHaveStyleRule(property: string, value: any, options?: StyleRuleOptions): R
    }
  }
}
```

The result: `tsc --noEmit` reported **487 `TS2339` errors** — every `toBeVisible`, `toBeInTheDocument`, `toHaveStyle`, `toHaveStyleRule` and friend. **The runtime was unaffected throughout: all 688 tests passed before and after.** The matchers are registered at runtime by `expect.extend`, which is untouched; only their type declarations went missing.

Both libraries are already at their latest published versions — jest-dom 7.0.1 (2026-08-09) and @emotion/jest 11.14.2 — and neither has a fix in progress:

| Library | Upstream issue | Opened | Last activity | State |
|---|---|---|---|---|
| `@testing-library/jest-dom` | [#738](https://github.com/testing-library/jest-dom/issues/738) — *"Vitest 5: matcher types are lost — `Assertion<T>` augmentation no longer merges with Vitest's `Assertion<R, T>`"* | 2026-09-06 | 2026-09-08 | Open. 3 comments, all from users; no maintainer reply, no label, no linked PR. Related: [#729](https://github.com/testing-library/jest-dom/issues/729), open since 2026-06-23 |
| `@emotion/jest` | [#3132](https://github.com/emotion-js/emotion/issues/3132) — *"Add vitest analog of @emotion/jest matchers without using jest types"* | 2023-11-28 | 2024-05-01 | Open, still labelled `needs triage` after ~3 years. 2 comments, both from users; no maintainer reply, no linked PR |

The two halves of the bridge therefore have very different outlooks. jest-dom's side is a bug reported four days ago against an active repository (last push 2026-08-09), affecting everyone migrating, and fixable in a few lines of its own `types/vitest.d.ts` — plausibly resolved upstream soon. emotion's side has been an untriaged feature request for three years, on a repository whose main branch has had no commits since 2025-11-04 and whose only recent move in this area was *widening* the peer range to `@types/jest@30.x`. **Treat the `toHaveStyleRule` half as indefinite rather than temporary.**

`src/vitest-matchers.d.ts` re-declares the matchers on vitest's own `Matchers` interface, which is the supported extension point:

```ts
declare module 'vitest' {
	interface Matchers<R extends void | Promise<void> = void, T = unknown>
		extends TestingLibraryMatchers<T, R> {
		toHaveStyleRule(property: string, value: unknown, options?: StyleRuleOptions): R;
	}
}
```

It reuses the libraries' own exported types rather than restating any matcher signature, apart from `toHaveStyleRule`, whose type emotion does not export separately.

This is also the approach the wider community is converging on: the workaround linked from jest-dom #738 is [a commit in owid-grapher](https://github.com/owid/owid-grapher/commit/d3b5d0efe2fc2ceb336bc66f87f076a9caa4a808) (2026-09-07) that extends vitest 5's `Matchers` with `TestingLibraryMatchers` in the same way, arrived at independently. The only alternative suggested in that issue is `@ts-expect-error` on every failing line — 487 of them here. **It is a bridge, to be deleted as soon as either library registers on vitest's `Matchers`** — the comment in the file says so. It stays out of the published surface: `tsconfig.lib.json` builds from `files: ["src/lib.ts", "src/testing.ts"]`, this file is a module imported by neither, and the api-extractor report comes out unchanged.

## The coverage change

Also fixed in 5: `include`/`exclude` globs were matched too eagerly. `**/types/*` now means what it says — files *directly* inside a `types/` directory — so nested paths are no longer excluded. One file joins the report:

```
src/types/network/soap.ts
```

It is not type-only: it exports the runtime type guard `isRawErrorSoapResponse`, re-exported through `src/types/network/index.ts` and `src/lib.ts`, so it is part of the public API. Vitest 4 silently left it unmeasured. **It arrives fully covered** (`FNF:1 FNH:1 LF:1 LH:1`), so the totals barely move — 81.22 → 81.23 statements, 71.15 → 71.18 functions.

I deliberately did **not** widen the glob to `**/types/**` to restore the old exclusion: that would go back to hiding runtime code from coverage. The remaining globs of the same shape (`**/tests/*`, `**/workers/*`) have no nested files today, so nothing else changed.

## Verification

Run locally on Node 22.23.2:

| Check | Result |
|---|---|
| `pnpm test` | ✅ 39 files, 688 tests passed |
| `pnpm test:ci` (junit + coverage, as CI runs it) | ✅ passed, `junit.xml` and `coverage/lcov.info` in place |
| `pnpm type-check` | ✅ clean |
| `pnpm lint` | ✅ 0 errors (5 pre-existing `sonarjs/cognitive-complexity` warnings) |
| `pnpm build:pkg` (`sdk build`) | ✅ compiled successfully |
| `pnpm build:lib` (api-extractor) | ✅ API report unchanged |
| Coverage file set vs vitest 4 | ✅ diffed: `soap.ts` is the only addition |

## Compatibility with #906

`vitest` 5 and `jsdom` 30 (#906) were also verified **together** on a throwaway merge of the two branches: 688 tests green and a clean type-check, so the two can merge in either order. As with the other open PRs, whichever merges last needs its `pnpm-lock.yaml` refreshed.

## Security

Resolves the 5 advisories that were reachable through `vitest` on the pre-update audit (`esbuild`, `vite`, `vitest`, `@vitest/mocker`).

## Noted, not acted on

Vitest 5 now prints a hint after each run: the suite spawns 39 isolated workers at ~657ms startup each, and `isolate: false` would save around 1.7s by reusing them across files. Worth trying, but it changes test isolation semantics, so it belongs in its own change rather than a dependency bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
 		"@types/react": "18.3.31",
 		"@types/react-dom": "18.3.7",
 		"@types/webpack-env": "1.18.8",
-		"@vitest/coverage-istanbul": "4.1.10",
+		"@vitest/coverage-istanbul": "5.0.0",
 		"@zextras/carbonio-ui-configs": "2.1.0",
 		"@zextras/carbonio-ui-sdk": "2.3.12",
 		"autoprefixer": "10.5.5",
@@ -117,7 +117,7 @@
 		"semantic-release": "25.0.9",
 		"sonarqube-scanner": "4.3.8",
 		"typescript": "5.9.3",
-		"vitest": "4.1.11",
+		"vitest": "5.0.0",
 		"vitest-fail-on-console": "0.10.1",
 		"webpack": "5.110.3",
 		"webpack-dev-server": "5.2.6"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -129,7 +129,7 @@ importers:
         version: 10.4.1
       '@testing-library/jest-dom':
         specifier: 7.0.1
-        version: 7.0.1(@testing-library/dom@10.4.1)(vitest@4.1.11)
+        version: 7.0.1(@testing-library/dom@10.4.1)(vitest@5.0.0)
       '@testing-library/react':
         specifier: 16.3.3
         version: 16.3.3(@testing-library/dom@10.4.1)(@types/react-dom@18.3.7(@types/react@18.3.31))(@types/react@18.3.31)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
@@ -155,8 +155,8 @@ importers:
         specifier: 1.18.8
         version: 1.18.8
       '@vitest/coverage-istanbul':
-        specifier: 4.1.10
-        version: 4.1.10(vitest@4.1.11)
+        specifier: 5.0.0
+        version: 5.0.0(vitest@5.0.0)
       '@zextras/carbonio-ui-configs':
         specifier: 2.1.0
         version: 2.1.0(@testing-library/dom@10.4.1)(@types/eslint@9.6.1)(typescript@5.9.3)
@@ -218,11 +218,11 @@ importers:
         specifier: 5.9.3
         version: 5.9.3
       vitest:
-        specifier: 4.1.11
-        version: 4.1.11(@opentelemetry/api@1.9.1)(@types/node@22.20.1)(@vitest/coverage-istanbul@4.1.10)(jsdom@29.1.1)(msw@2.15.0(@types/node@22.20.1)(typescript@5.9.3))(vite@7.3.2(@types/node@22.20.1)(jiti@2.7.0)(less@4.6.7)(terser@5.51.2))
+        specifier: 5.0.0
+        version: 5.0.0(@opentelemetry/api@1.9.1)(@types/node@22.20.1)(@vitest/coverage-istanbul@5.0.0)(jsdom@29.1.1)(msw@2.15.0(@types/node@22.20.1)(typescript@5.9.3))(vite@7.3.2(@types/node@22.20.1)(jiti@2.7.0)(less@4.6.7)(terser@5.51.2))
       vitest-fail-on-console:
         specifier: 0.10.1
-        version: 0.10.1(@vitest/utils@4.1.11)(vite@7.3.2(@types/node@22.20.1)(jiti@2.7.0)(less@4.6.7)(terser@5.51.2))(vitest@4.1.11)
+        version: 0.10.1(@vitest/utils@4.1.11)(vite@7.3.2(@types/node@22.20.1)(jiti@2.7.0)(less@4.6.7)(terser@5.51.2))(vitest@5.0.0)
       webpack:
         specifier: 5.110.3
         version: 5.110.3(postcss@8.5.15)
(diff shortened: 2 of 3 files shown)


lisaparma (dismissed):

lisaparma (approved):

CataldoMazzilli:
Rebased on the current `main` and moved the pair to **5.0.1** (released 2026-09-15), so this branch is mergeable again.

- The conflict was confined to `package.json` and `pnpm-lock.yaml`. Resolved by taking `main` for every dependency — including jsdom 30.1.0, ui-configs 2.1.1 and the typescript-eslint v8 fixes — and keeping only `vitest` and `@vitest/coverage-istanbul` from this branch; the lockfile was regenerated from `main`'s.
- **`src/vitest-matchers.d.ts` is byte-identical.** 5.0.1 emits the same `Matchers` declaration as 5.0.0, so the bridge needed no change and `tsc --noEmit` is clean.
- Full suite re-run against the new base: 39 files / 688 tests green, lint 0 errors, `sdk build` and api-extractor fine, API report unchanged. Coverage re-diffed against the current `main`: 101 → 102 files, `src/types/network/soap.ts` still the only addition, fully covered.
- Upstream is unchanged: jest-dom is still on 7.0.1 with [#738](https://github.com/testing-library/jest-dom/issues/738) open and unanswered by the maintainers (a [second reporter](https://github.com/testing-library/jest-dom/issues/738#issuecomment-5634263560) confirmed this same `Matchers` shape on 2026-09-11 and offered to PR it), and @emotion/jest [#3132](https://github.com/emotion-js/emotion/issues/3132) has not moved since 2024. Staying in draft until at least the jest-dom half lands.

Sorry for the dismissed approval — the push invalidated it again.

